PR #11, opened 2019-03-09
Add support for python 3.4

get9:
`contextlib.redirect_stderr/stdout` was introduced in python 3.4. This
change adds extra restrictions to the redirect_stderr/stdout
redefinitions as well as adds CI for testing on py34 and py35.

Also forces pip install of black for any python version less than 3.6,
as black is only officially supported on >= 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,23 @@
 language: python
 python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-  - "3.7"
+  - 2.7
+  - 3.4
+  - 3.5
+  - 3.6
 install:
   - make install
 script:
   make
-dist: xenial
-sudo: true
+
+matrix:
+  include:
+    # workaround for https://github.com/travis-ci/travis-ci/issues/9815
+    - python: 3.7
+      dist: xenial
+      sudo: true
+
+    # black requires 3.6
+    - name: "Black formatting"
+      python: 3.6
+      install: pip install black
+      script: black --check testslide/ tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
+  - "3.5"
   - "3.6"
   - "3.7"
 install:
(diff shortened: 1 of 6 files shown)


get9:
@fornellas: Getting an error with one particular unit test, mind taking a look? I'm not sure why this would fail only on 3.4.

.travis.yml
```diff
@@ -1,11 +1,23 @@
...
+matrix:
+  include:
+    # workaround for https://github.com/travis-ci/travis-ci/issues/9815
+    - python: 3.7
+      dist: xenial
+      sudo: true
+
+    # black requires 3.6
+    - name: "Black formatting"
+      python: 3.6
```
fornellas:
```suggestion
      python: 3.7
```

fornellas: Can't this be latest (3.7)?

Makefile
```diff
@@ -8,7 +8,7 @@ all: test
 
 .PHONY: black_check
 black_check:
-	if python -c 'import sys ; sys.exit(1 if sys.version.startswith("2.") else 0)' ; then black --check testslide/ tests/ ; fi
+	if command -v black; then black --check testslide/ tests/ ; fi
```
fornellas: There was already a conditional to require black for supported python versions at setup.py, plus this conditional here, any reason we can't keep that?

testslide/mock_callable.py
```diff
@@ -66,8 +66,8 @@ def _format_args(indent, *args, **kwargs):
     s += "{"
     if kwargs:
         s += "\n"
-        for k, v in kwargs.items():
-            s += "{}  {}={},\n".format(indentation, k, v)
+        for k in sorted(kwargs.keys()):
```
fornellas: Nice touch!

.travis.yml
get9: I suppose it can, but I didn't want to rely on 3.7 since you need the Ubuntu hack to get it working. I can make it 3.7 if you prefer.

Makefile
get9: Writing the conditional to support black requires checking the python version is at least 3.6. It got to be a lot longer to fully write out that check, so I just decided to check and see if black was installed. That seems like a more reasonable check, plus we ensure on CI that it's installed. I would argue if the user has black provided via some other means, you probably want this target to succeed.

PR #14, opened 2019-04-15
[Simple] Update index.rst for a non-existing call

xsh6528:


fornellas:
https://testslide.readthedocs.io/en/1.4.9/mock_callable/index.html > new release with the fix, thanks!

PR #17, opened 2019-04-18
Implemented CallerInfoException to display additional info with Exception raises

xteodora:
Implemented CallerInfoException to display caller file name and line when exceptions are raised.

Also added a new assert assertRaisesWithRegexMessage so that we will be able to test the new message given that the string will begin with " .... {file_path}:{line no} ..."

### DIFF
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -12,6 +12,7 @@ import sys
 import dis
 import copy
 import functools
+import inspect
 
 if sys.version_info[0] >= 3:
     from unittest.mock import create_autospec, _must_skip
@@ -187,6 +188,7 @@ class StrictMock(object):
         self.__dict__["__template"] = template
         self.__dict__["__runtime_attrs"] = runtime_attrs or []
         self.__dict__["__name"] = name
+        self.__dict__["__caller"] = inspect.getframeinfo(inspect.stack()[1][0])
 
         if (
             self.__template
@@ -323,8 +325,12 @@ class StrictMock(object):
             name = " name={}".format(repr(self.__dict__["__name"]))
         else:
             name = ""
-        return "<StrictMock 0x{:02X}{name}{template}>".format(
-            id(self), name=name, template=template
+        return "<StrictMock 0x{:02X}{name}{template} - {filename}:{lineno}>".format(
+            id(self),
+            name=name,
+            template=template,
+            filename=self.__dict__["__caller"].filename,
+            lineno=self.__dict__["__caller"].lineno,
         )
 
     def __get_copy(self):
--- a/tests/strict_mock_testslide.py
+++ b/tests/strict_mock_testslide.py
@@ -14,6 +14,7 @@ import contextlib
 import copy
 import functools
 import sys
+import re
 
 from testslide.dsl import context, xcontext, fcontext, Skip  # noqa: F401
 
@@ -97,12 +98,33 @@ def strict_mock(context):
             ),
         )
 
+    @context.function
+    @contextlib.contextmanager
+    def assertRaisesWithRegexMessage(self, exception, rgx):
+        with self.assertRaises(exception) as cm:
+            yield
+        ex_msg = str(cm.exception)
+
+        if not re.search(rgx, ex_msg):
+            self.assertEqual(
+                ex_msg,
+                rgx,
+                "Expected exception {}.{} message "
+                "to match regex\n{}\nbut got\n{}.".format(
+                    exception.__module__, exception.__name__, repr(rgx), repr(ex_msg)
+                ),
+            )
+
     @context.sub_context
     def without_template(context):
         @context.before
         def before(self):
             self.strict_mock = StrictMock()
-            self.strict_mock_str = "<StrictMock 0x{:02X}>".format(id(self.strict_mock))
+            self.strict_mock_rgx = (
+                "<StrictMock 0x{:02X} - ".format(id(self.strict_mock))
+                + re.escape(__file__)
+                + ":\d+>"
+            )
             self.value = 2341234123
 
         @context.example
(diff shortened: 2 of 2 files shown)


tests/mock_callable_testslide.py
```diff
@@ -104,6 +105,23 @@ def assertRaisesWithMessage(self, exception, msg):
...
         )
 
+    @context.function
+    @contextlib.contextmanager
+    def assertRaisesWithRegexMessage(self, exception, rgx):
+        with self.assertRaises(exception) as cm:
+            yield
+        ex_msg = str(cm.exception)
+
+        if not re.search(rgx, ex_msg, re.IGNORECASE):
```
fornellas: I'd remove IGNORECASE here.

tests/mock_callable_testslide.py
```diff
@@ -193,7 +211,14 @@ def it_rejects_unknown_arguments(self):
                                             self.specific_call_kwargs.keys()
                                         )
                                     )
-                                    + "    }\n",
+                                    + "    }\n"
+                                )
+                                rgx = (
+                                    "The mock_callable configured at ((?:[^/]*/)*)(.*):\d{0,4} encountered unexpected behavior:\n"
```
fornellas: I believe we can get significant more coverage here if we use [``__file__``](https://docs.python.org/3/reference/import.html#__file__) instead of `((?:[^/]*/)*)(.*)`. Making this match the line on a case by case would be a bit of a pain to keep up to date, I wouldn't bother with that.

fornellas:
This is overall legit, but I see some issues. To make things easier, I just [added more thorough tests for StrictMock](https://github.com/facebookincubator/TestSlide/commit/556229c7ca4ccb8421a19784b14a94ae348e5e3f), which you should rebase on top, and I believe will expose some bugs (see below).

Looking at the exceptions:

```
StrictMock
	UndefinedBehavior
	NoSuchAttribute
mock_callable()
	UndefinedBehaviorForCall
	UnexpectedCallReceived
	UnexpectedCallArguments
```

They don't seem to be shareable across StrictMock and mock_callable(), perhaps it is worth making this clear at their descriptions and messages that were changed.

I believe that `inspect.getframeinfo(inspect.stack()[1][0])` won`t properly get the correct place where the call was created, but rather, somewhere inside TestSlide itself for most cases. One way out I can think of is:

1 - Get the call stack.
2 - Go up the stack, until you find the first non-TestSlide line.
3 - Use it.

(2) can be a bit tricky, because the tests... are part of TestSlide, so lines that match TestSlide tests module should be considered "outside". This logic should probably go in a shared function (some new `lib.py`?) and use it at:

- At `StrictMock.__init__`, and refer the values at `StrictMock.__repr__` (this should cover all cases).
- At `_MockCallableDSL.__init__`, and refer the values in all places exceptions are created at the same file.

Note that `_MockCallableDSL` is the engine below `mock_constructor` and `mock_callable`, so perhaps, either name should come to `_MockCallableDSL.__init__`, so we can have proper exception messages.

Makes sense?

testslide/strict_mock.py
```diff
@@ -268,21 +270,27 @@ def __setattr__(self, name, value):
...
             # doesn't event exist in the template class
             if hasattr(self.__template, name):
+                caller = inspect.getframeinfo(inspect.stack()[1][0])
                 raise UndefinedBehavior(
                     self,
                     name,
                     "The attribute {} is defined in the template class "
                     "{}, but its behavior is not yet defined in this "
                     "StrictMock".format(self.__template_name),
+                    # caller.filename,
```
fornellas: This can be dropped.

testslide/strict_mock.py
```diff
@@ -293,11 +301,14 @@ def __getattr__(self, attr):
                 pass
 
         if self.__can_mock_attr(attr):
+            caller = inspect.getframeinfo(inspect.stack()[1][0])
```
fornellas: I believe this can go away.

testslide/strict_mock.py
```diff
@@ -268,21 +270,27 @@ def __setattr__(self, name, value):
             # behavior we use a different exception than when the attribute
             # doesn't event exist in the template class
             if hasattr(self.__template, name):
+                caller = inspect.getframeinfo(inspect.stack()[1][0])
```
fornellas: I don't think this is in use anymore.

testslide/strict_mock.py
```diff
@@ -268,21 +270,27 @@ def __setattr__(self, name, value):
...
                     self,
                     name,
                     "The attribute {} is defined in the template class "
                     "{}, but its behavior is not yet defined in this "
                     "StrictMock".format(self.__template_name),
+                    # caller.filename,
+                    # caller.lineno,
                 )
             else:
+                caller = inspect.getframeinfo(inspect.stack()[1][0])
```
fornellas: I don't think this is in use anymore.

testslide/strict_mock.py
```diff
@@ -268,21 +270,27 @@ def __setattr__(self, name, value):
...
                 raise NoSuchAttribute(
                     self,
                     name,
                     "Can not set attribute {} that is neither "
                     "part of template class {} or runtime_attrs={}.".format(
                         name, self.__template_name, self.__runtime_attrs
                     ),
+                    # caller.filename,
```
fornellas: To drop.

testslide/strict_mock.py
```diff
@@ -293,11 +301,14 @@ def __getattr__(self, attr):
                 pass
 
         if self.__can_mock_attr(attr):
+            caller = inspect.getframeinfo(inspect.stack()[1][0])
             raise UndefinedBehavior(
                 self,
                 attr,
                 "Can not getattr() an undefined StrictMock "
                 "attribute. Use setattr() to define it.",
+                # caller.filename,
```
fornellas: To drop.

fornellas:
Thanks @xteodora for making this happen!

PR #22, opened 2019-06-11
Drop usage of inspect2

fornellas:
There's no hard need to require inspect2, let's drop this requirement.


fornellas:
Dropped Python 3.4, it reached EOL https://devguide.python.org/devcycle/#end-of-life-branches.

PR #26, opened 2019-06-21
Fix mock_* assertions with DSL

fornellas:
Assertions for `mock_callable()` / `mock_constructor()` are not working as expected for all cases when using the DSL.

Due to an implementation bug, assertions are executed for the last context declared, not the context that was actually executed. Tests were passing, because they were not exercising this more complex scenario. Integration with Python`s Unittest is not affected.

This PR adds test coverage for the DSL +  `mock_callable()` / `mock_constructor()` integration, that fails without the fixes here.

### DIFF
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -268,9 +268,6 @@ class Example(object):
         for after in reversed(after_functions):
             with aggregated_exceptions.catch():
                 after(context_data)
-        for assertion in self.context.assertions:
-            with aggregated_exceptions.catch():
-                assertion()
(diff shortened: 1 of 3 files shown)


testslide/__init__.py
```diff
@@ -184,7 +184,10 @@ def raise_correct_exception(self):
         if Skip in ex_types or unittest.SkipTest in ex_types:
             raise Skip()
         else:
```
lucasvasconcelos:
```suggestion
        elif len(self.exceptions) == 1:
            raise self.exceptions[0]
        else:
            raise self
```

PR #30, opened 2019-06-24
Simplify Python 3.7 build

fornellas:
Hopefully we don't need the old hack anymore.


shish:
Should this be rebased on top of the "make black an official dep" branch instead of duplicating it in this branch?

fornellas:
Rebased.

PR #31, opened 2019-06-24
Fix assertion on call limit

fornellas:
A regression was recently introduced, that broke `.and_assert_not_called()` when you have mock_callable setup for different call arguments such as:

```python
self.mock_callable(target, 'attr').to_return_value(None)
self.mock_callable(target, 'attr').for_call(1, 2).and_assert_not_called()
```

Previously, this would break:

```python
target.attr("other", "arguments")
```

but it must pass. This PR fixes that.

### DIFF
--- a/testslide/mock_callable.py
+++ b/testslide/mock_callable.py
@@ -116,9 +116,16 @@ class _Runner(object):
         if self.max_calls and self.call_count > self.max_calls:
             raise UnexpectedCallReceived(
                 (
-                    "{}, {}: Unexpected call received.\n"
-                    "  Expected to receive at most {} calls, but an extra call was made."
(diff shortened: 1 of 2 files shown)


testslide/mock_callable.py
```diff
@@ -659,16 +666,14 @@ def and_assert_called_exactly(self, count):
         If assertion is for 0 calls, any received call will raise
         UnexpectedCallReceived and also an AssertionError.
         """
-        if not count:
-            self._runner = None
```
fornellas: We can't simply drop the existing runner, as it takes care of the accepted arguments. _Runner already implements the upper limit on calls, we don't need to repeat ourselves here, just ensure there's a runner if there's none.

PR #32, opened 2019-06-25
Fix mock_constructor() and with_wrapper()

fornellas:
This PR fixes 2 stacked bugs:

- `with_wrapper()` was failing with `TypeError: original_callable() takes at least 1 argument (0 given)`.
- Once this was fixed, it exposed a tricky underlying bug, that would make it raise `TypeError: super(type, obj): obj must be an instance or subtype of type`. This happened exclusively when `__init__` of the mocked class called `super(SomeClass, self)`. In this case, `type(self)` would be `SomeClass` (the original) and `SomeClass` would resolve to `SomeClassMock` (that `mock_constructor()` puts there), making `super()` rightfully blow up.

Closes facebookincubator/TestSlide/issues#15.

### DIFF
--- a/tests/mock_constructor_testslide.py
+++ b/tests/mock_constructor_testslide.py
@@ -14,10 +14,15 @@ import contextlib
 from testslide.dsl import context, xcontext, fcontext, Skip  # noqa: F401
 
 
-class Target(object):  # noqa
+class BaseTarget(object):
     def __init__(self, *args, **kwargs):
         self.args = args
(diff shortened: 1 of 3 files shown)


fornellas:
The with_wrapper() issue will be fixed by https://github.com/facebookincubator/TestSlide/pull/33. The super() issue will be fixed in a future PR, as the fix here won't work for methods other than `__init__`.

Issue #15, opened 2019-04-16
`with_wrapper` does not support `mock_constructor`


xsh6528:
```
from testslide.dsl import context


class MyClass(object):
    __slots__ = ["value"]

    def __init__(self, value):
        self.value = value


@context  # noqa
def mock_constructor_with_wrapper_test(context):

    @context.sub_context
    def intercept_context(context):
        @context.before
        def intercept_call_to_constructer(self):
            def wrapper(original_callable, *args, **kwargs):
                obj = original_callable(*args, **kwargs)
                obj.value += 1
                return obj

            self.mock_constructor(
                __name__, "MyClass"
            ).with_wrapper(wrapper).and_assert_called_once()

        @context.example
        def call_constructor(self):
            obj = MyClass(value=1)
            self.assertEqual(obj.value, 2)

    @context.example
    def call_constructor(self):
        obj = MyClass(value=1)
        self.assertEqual(obj.value, 1)

```

Error

```
Failures:

  0) mock constructor with wrapper test, intercept context: call constructor
    1) TypeError: original_callable() takes at least 1 argument (0 given)
      File "experimental/shihaoxu/testslide/mock_constructor_with_wrapper_test.py", line 31, in call_constructor
        obj = MyClass(value=1)
      File "experimental/shihaoxu/testslide/mock_constructor_with_wrapper_test.py", line 21, in wrapper
        obj = original_callable(*args, **kwargs)
```

fornellas:
It seems the issue is [here](https://github.com/facebookincubator/TestSlide/blob/master/testslide/mock_constructor.py#L85): the `_` should likely be dropped.

How do you feel about doing a PR for this, also add your awesome test case to reproduce the issue [here](https://github.com/facebookincubator/TestSlide/blob/master/tests/mock_constructor_testslide.py)?

danielkza:
Looking a bit more into this, it seems simply dropping the `_` in the callback is not sufficient. In my understanding the problem is rooted in the nature of `mock_constructor`: while the intended use case is to configure it as a constructor call (that *does not* take the class as the first argument), in practice it is setup as a mock of the `__new__` method (which *does* take the constructed class as the first argument). If we do the simple fix we can fix one usage but we break the other.

To deal with that "duality", adding a new subclass of `_MockCallableDSL` that properly deals with the argument seems like a good approach. That can substitute the `prepend_first_arg` machinery in a less error-prone way.

fornellas:
[...]

PR #33, opened 2019-06-25
Refactor mock_constructor to better support all mocking features

danielkza:
Previously mock_constructor simply created a regular mock for the
__new__ method of its target. Unfortunately that causes issues when
using more complex mocking features, since users expect to pass
arguments as they would to constructor (without the class as the first
argument).

To fix that:

- Introduce _ConstructorRunner, which can wrap any other runner for Mock
  actions while ensuring the initial class argument is not passed on,
  and corresponds to a mock class generated by TestSlide.
- Introduce _MockConstructorDSL, wrapping methods which need special
  accounting for constructor arguments, and ensuring all runners are
  wrapped by a _ConstructorRunner

Fixes: #15

### DIFF
--- a/tests/mock_constructor_testslide.py
+++ b/tests/mock_constructor_testslide.py
@@ -12,6 +12,7 @@ import sys
 import contextlib
 
 from testslide.dsl import context, xcontext, fcontext, Skip  # noqa: F401
+from testslide.mock_callable import _MockCallableDSL
 
 
 class Target(object):  # noqa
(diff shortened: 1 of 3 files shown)


fornellas:
I really like the idea of `_ConstructorRunner`, it seems to be a much cleaner solution than how it currently is (and what https://github.com/facebookincubator/TestSlide/pull/32 builds on top).

Only thing that concerns me, is test coverage. Python internals is a mine field: you fix one thing, another breaks... mock_callable has a quite extensive test coverage, and mock_constructor kinda piggy back on that coverage, and there are only some smoke tests (as the ones you added here). `_ConstructorRunner` has some extra logic, which we might not totally cover.

I'm just wondering, if we keep `_MockConstructorDSL._add_runner`, and simplify the footprint of `_ConstructorRunner`, might be possible. Eg: dynamically subclass the given `_Runner` subclass, and add the extra logic needed for the extra argument. What do you think?

Also, I *think* the tests I added at https://github.com/facebookincubator/TestSlide/pull/32 might expose the super issue I mentioned, which still needs fixing.

fornellas:
So, I suppose we can merge this PR instead of https://github.com/facebookincubator/TestSlide/pull/32. Do you think you can pull some changes from https://github.com/facebookincubator/TestSlide/pull/32 here?

- Add the `BaseTarget` changes, to trigger the `super()` issue, and also add the fix (`original_callable()` changes).
- Confirm that the tests here also cover the same scenarios as [here](https://github.com/facebookincubator/TestSlide/pull/32/files#diff-16350229bc78bcd6128ef98ae861eb8cR105).

Thanks!

PR #34, opened 2019-06-28
Fix mock_constructor() patching

fornellas:
The way `mock_constructor()` patching works, it breaks the behavior of `super()` when called from inside any methods, whenever `with_wrapper()` is used:

```python
super(Target, self)
```
will always fail, because after the patching, `Target` will be a subclass of `Target` (`TargetMock`), and `self`, an instance of `Target`, thus giving us:

```
`TypeError: super(type, obj): obj must be an instance or subtype of type`.
```

This PR changes the strategy used for patching:

- It creates a subclass of `Target`.
- Copy all attributes from `Target` to the new subclass.
- Patch the new subclass in place of the original class.

Things being like this, any `super` calls will always work. It comes though, with some issues:

- After `__new__` is called, the interpreter unconditionally calls `__init__`, with the default arguments. Thanks to @danielkza new test coverage, this issue was surfaced, and a mechanism was added to call `__init__` only once with the correct arguments.

There are a few things missing to fully wrap up this issue:

- Potentially add a safety check: if there's more than 1 reference to the class being patched, as there's no guarantee things would work in this scenario.
- Make sure this works with metaclasses (add tests).

CC @lucasvasconcelos

### DIFF
--- a/docs/mock_constructor/index.rst
+++ b/docs/mock_constructor/index.rst
@@ -70,7 +70,7 @@ mock_constructor is a way to not only solve this for Python 3, but also provide 
 Internally, mock_constructor will:
 
 * Patch the class at its module with a subclass of it, that is dynamically created.
-* ``__new__`` of this dynamic subclass is handled by mock_callable.
+* This new subclass is essentially a copy of the original class, but overrides its ``__new__`` with a factory that handles `mock_constructor()` interface.
 
 Integration With Other Frameworks
 ---------------------------------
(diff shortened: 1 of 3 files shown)


fornellas:
@danielkza I made the tests a bit saner https://pastebin.com/cQZmPNU2. There's a skipped test case for `isinstance` which apparently depends on the metaclass mechanics which I could use a hand (potentially a PR on top). Thanks!

PR #36, opened 2019-07-15
Make mock_callable() refuse non callable targets

fornellas:
`mock_callable()` only works (and is only tested) to work with functions and methods. Let's give a fast failure, in case some other target is given, like a non callable attribute or a class.

### DIFF
--- a/testslide/mock_callable.py
+++ b/testslide/mock_callable.py
@@ -423,13 +423,13 @@ def _patch(target, method, new_value):
         original_callable = getattr(target, method)
         if not callable(original_callable):
             raise ValueError(
-                "mock_callable() can only be used with callable attributes, got: {}".format(
+                "mock_callable() can only be used with callable attributes and {} is not.".format(
                     repr(original_callable)
                 )
             )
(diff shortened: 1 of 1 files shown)


testslide/mock_callable.py
```diff
@@ -421,6 +421,18 @@ def _patch(target, method, new_value):
         original_callable = None
     else:
         original_callable = getattr(target, method)
+        if not callable(original_callable):
+            raise ValueError(
+                "mock_callable() can only be used with callable attributes and {} is not.".format(
+                    repr(original_callable)
+                )
+            )
+        if type(original_callable) is type:
```
shish: Does this work right for python 2? The error message suggests that we're trying to reject classes, but old-style python2 classes are still allowed:

```
>>> class foo(object):
...     pass
...
>>> type(foo) is type
True
```

```
>>> class bar:
...     pass
...
>>> type(bar) is type
False
```

fornellas:
Fair point, changed to use `inspect.isclass()` that handles such case.

Issue #19, opened 2019-04-23
Support for async functions with mock_callable()


fornellas:
Support for [async functions](https://docs.python.org/3/library/asyncio.html) with `mock_callable()`.

Here's a [half working patch](https://github.com/fornellas/TestSlide/commit/e157bbf06c4ff0c6ed9264f00e3610665c395196) for that, as a starting point.

To prove that any solution works, all relevant tests at `tests/mock_callable_testslide.py` must be updated to to cover async cases (eg: async function at module, async instance method, async class method etc).

PS: previous PR https://github.com/facebookincubator/TestSlide/pull/8

david-caro:
Hi @fornellas, I have a question about this, in order to support asyncio, the whole module would have to drop python2 support right?

fornellas:
I suppose you're correct. Python 2 is dying anyway, it's support can be dropped at this point.

fornellas:
PS: I created a milestone for async stuff https://github.com/facebookincubator/TestSlide/milestone/1, as there's more stuff to implement.

jivid:
I've put up #60 to provide `mock_async_callable`. I do believe having an explicit `mock_async_callable` is better than having the `mock_callable` interface detect whether it's received a coroutine or a regular function. The latter just feels like it would hide too much from the users.

fried:
unittest.IsolatedAsyncioTestCase in 3.8 uses that.  So to have async test methods they have to explicitly be coroutines.  No sync methods that return awaitables, since that's pretty hairy. 
Also the AsyncMock in 3.8 is also a really good interface for async mocks, in that it looks at the spec and if its a coroutine the mocks that return when that is called is another AsyncMock,  otherwise you get a normal MagicMock.

PR #53, opened 2019-10-17
[global] Drop python2 support

david-caro:
This will simplify and allow to add support for async.
Python2 is out of maintenance in 2020.

THIS IS UNTESTED, DO NOT MERGE YET


fornellas:
@david-caro my apologies, it seems we both went after dropping Python 2 at the same time, I just published #55 and only then found out about your PR.

I compared both, it seems that we're mostly covering the same ground, with a few differences:

- Your PR contains a lot of codemods (eg: drop classes inheriting from object).
- You dropped some p2/p3 style super calls from the tests. I think we still need those, as Python 2 style super calls are still valid, and there are a lot of corner cases there.
- It seems there's some async stuff at the PR that are part of #19 .

How do you feel about rebasing this PR on top of #55 , push all codemods you have here, and let the async stuff for another PR at #19 ?

david-caro:
I'm going on a 3 week long holiday today, so I will not be able to work on this for a while, feel free to close and take over whatever you think it's useful and drop the rest.

fornellas:
Will do, thank you! Apologies again for the race here.

fornellas:
@david-caro I just merged #55  and tired rebasing this on top. It seems the nice takeaway is the import ordering. I'd love to have the import ordering lint as part of the build process [here](https://github.com/facebookincubator/TestSlide/blob/master/Makefile). I don't have experience with that, but it seems there's a [flake8 plugin](https://github.com/PyCQA/flake8-import-order) for that. Do you think we can get this to work, so we have a permanent solution for import ordering?

Issue #57, opened 2019-10-18
Async support for the DSL


fornellas:
Add async support for DSL:

```python
@context.example
async def async_example(self):
  pass

@context.before
async def async_before(self):
  pass

@context.after
async def async_after(self):
  pass

@context.around
async def async_around(self, example):
  example()

@context.memoize
async def async_memoize(self):
  pass

@contex.function
async def asyc_function(self):
  pass
```

PR #59, opened 2019-10-25
Add async support for DSL

fornellas:
Add support for [TestSlide's DSL](https://testslide.readthedocs.io/en/1.7.0/testslide_dsl/index.html) to test async stuff.

Example usage:

```python

@context
def testing_async_code(context):
  @context.around
  async def around(self, example):
    await example()  # Note that this must be awaited!

  @context.before
  async def before(self):
    pass

  @context.after
  async def after(self):
    pass

  @context.memoize_before
  async def memoize_before(self):
    return "memoize_before"

  @context.function
  async def function(self):
    return "function"

  @context.example
  async def example(self):
    assert self.memoize_before == "memoize_before"
    assert self.function == "function"
```

There's some ugliness at `_ExampleRunner`: `def _sync_run_all_hooks_and_example` and `async def _real_async_run_all_hooks_and_example` implement the exact same thing, one being sync the other async. I left big warnings at both to keep them in sync. Unfortunately I found no better way of sharing logic between sync and async code. I added a smoke test for the async version of the function, that should cover most of the ground to mitigate that.

Closes of #57 .

### DIFF
--- a/docs/testslide_dsl/async_support/index.rst
+++ b/docs/testslide_dsl/async_support/index.rst
@@ -1,12 +1,15 @@
 Async Support
 =============
 
-TestSlide's DSL supports `asynchronous I/O <https://docs.python.org/3/library/asyncio.html>`_ testing. For that you must:
+TestSlide's DSL supports `asynchronous I/O <https://docs.python.org/3/library/asyncio.html>`_ testing.
 
-- Declare contexts as sync functions as usual.
-- Declare examples and hooks (around, before and after) as async.
+For that, you must declare all of these as async:
 
-Example:
+- Hooks: around, before and after.
+- Examples.
+- Memoize before.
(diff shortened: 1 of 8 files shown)


jivid:
Hah I implemented a slightly different version of this in #60 just now :) Want to have a look at that approach and see how to proceed?

fornellas:
@jivid I managed to make this work by either supporting all example and hook (before, after & around) code to be either sync or async. What do you think?

I'll update the PR with 3.6 fix, docs and test cases for failures when miking sync+async stuff, but this seems to cut it.

fornellas:
Ready for review! Managed to make everything work with proper documentation.

fornellas:
Closed by mistake, #68 reissues this.

PR #60, opened 2019-10-27
Add mock_async_callable for asyncio support

jivid:
Add a `mock_async_callable` DSL which works exactly like `mock_callable` but for `async` functions.

The decision to not patch `mock_callable` to support `async` functions was intentional and made to make it clear when reading the test what kind of code is being mocked. Just as you can't just stick `async` in front of a function and expect everything to work, you shouldn't rely on `mock_callable` just working either.

### DIFF
--- a/testslide/mock_callable.py
+++ b/testslide/mock_callable.py
@@ -20,6 +20,15 @@ def mock_callable(target, method):
     return _MockCallableDSL(target, method)
 
 
+def mock_async_callable(target, method):
+    if method == "__new__":
+        raise ValueError(
+            "Mocking __new__ is not allowed with mock_callable(), please use "
+            "mock_constructor()."
+        )
+    return _MockAsyncCallableDSL(target, method)
+
+
 _unpatchers = []  # type: List[Callable]  # noqa T484
 
 
@@ -411,6 +420,16 @@ class _CallableMock(object):
         return [runner.accepted_args for runner in self.runners if runner.accepted_args]
 
 
+class _CallableAsyncMock(_CallableMock):
+    def __call__(self, *args, **kwargs):
+        _super = super(_CallableAsyncMock, self)
+
+        async def coro_call():
+            return _super.__call__(*args, **kwargs)
+
+        return coro_call()
+
+
 class _DescriptorProxy(object):
     def __init__(self, original_class_attr, attr_name):
         self.original_class_attr = original_class_attr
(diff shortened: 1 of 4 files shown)


testslide/mock_callable.py
```diff
@@ -20,6 +20,15 @@ def mock_callable(target, method):
     return _MockCallableDSL(target, method)
 
 
+def mock_async_callable(target, method):
+    if method == "__new__":
+        raise ValueError(
+            "Mocking __new__ is not allowed with mock_callable(), please use "
+            "mock_constructor()."
+        )
+    return _MockAsyncCallableDSL(target, method)
```
fornellas: The logic for `__new__` is duplicated, it could be moved to `_MockCallableDSL`.

fornellas:
Neat! This seems in a good direction! There are a few things that we need to look over as well:

- [Documentation](https://github.com/facebookincubator/TestSlide/blob/master/docs/mock_callable/index.rst) needs to be updated to explain about async usage (including test framework integration) . Tip: do `make docs` to build and test locally.
- `mock_callable()` supports [a lot of different configurations](https://testslide.readthedocs.io/en/1.7.0/mock_callable/index.html#cheat-sheet). I don't expect them all to work with the current implementation, we need tests for all of them.

From the top of my mind, I do expect these to "just work":

- `.to_return_value(value)`
- `.to_return_values(values_list)`
- `.to_raise(exception)`
- `.and_assert_called_exactly(times)`
- `.and_assert_called_once()`
- `.and_assert_called_twice()`
- `.and_assert_called_at_least(times)`
- `.and_assert_called_at_most(times)`
- `.and_assert_called()`
- `.and_assert_called_ordered()`
- `.and_assert_not_called()`

But I have my doubts about these:

- `.for_call(*args, **kwargs)`
- `.to_yield_values(values_list)`
- `.with_implementation(func)`
- `.with_wrapper(func)`
- `.to_call_original()`

Unfortunately Python has shown to be full of quicks and exception cases on how things work. The [current tests](https://github.com/facebookincubator/TestSlide/blob/async_dsl/tests/mock_callable_testslide.py), while *hard to maintain and confusing* (sorry about that), were extremely helpful in getting all corner cases in the past, we'd loose a lot if we did not have the same coverage for async.

Ideally, we'd need to duplicate all sync wyth async equivalents of the tests, but in some cases (eg: iterator, calling original) we'd need more async specific stuff. I'm happy to assist you on that if needed.

jivid:
Thanks for the comments @fornellas, I'll update this PR with the requested changes

jivid:
@fornellas  I'm still working on this, not ready for review yet. Just pushed my changes so far so I don't lose them.

fornellas:
FYI, #59 should help testing async stuff, I'm still working on it.

fornellas:
@jivid as we spoke, I'm gonna start working on this issue now so we can deliver it ASAP. I'm closing this PR so we don't clash, but feel free to ping me to join forces if you can spare bandwidth.

Thanks for what we have here already, I have something to grow on top :-D

PR #62, opened 2019-10-30
Drop Python 3.5 support

fornellas:
Let's not spend time supporting this 4 years old branch and get rid of not being able to use 3.6+ goodies.

### DIFF
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -15,31 +15,11 @@ import testslide.mock_callable
 import testslide.mock_constructor
 from testslide.strict_mock import StrictMock  # noqa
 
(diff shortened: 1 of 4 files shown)


testslide/__init__.py
```diff
@@ -15,31 +15,11 @@
...
-        sys.stderr = target
-        try:
-            yield
-        finally:
-            sys.stderr = original
-
-
-if sys.version_info[0] < 3:
-    raise RuntimeError("Python >=3 required.")
+if sys.version_info[0] < 3 or (
```
pmazzini: Can't this be changed for `sys.version_info >= (3, 6)`?

fornellas: Neat: `sys.version_info >= (3, 6)`, learned one today!

PR #63, opened 2019-10-30
Refactor CLI tests

fornellas:
The CLI tests were executing the CLI interface on the same process, thus running not only the sample tests, but all tests declared on the same file. This is creating problems when implementing async support, as the test runner is running another test runner inside itself: an event loop inside an event loop.

This diff refactor the tests:

- All sample tests were moved to `sample_dsl.py`.
- All CLI tests now spawn an independent Python process and captures its output.
- Assertions were tightened a bit to ensure only the expected tests are in fact executed.

### DIFF
--- a/tests/cli_unittest.py
+++ b/tests/cli_unittest.py
@@ -3,25 +3,14 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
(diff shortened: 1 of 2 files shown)


fornellas:
3.5 failures can be ignore: #62 is dropping it.

PR #66, opened 2019-11-04
Refactor strict mock

### LINKED ISSUE #23: Fix StrictMock with bool()
Given:

```python
In [1]: from testslide import StrictMock                                      
In [2]: class A: 
   ...:     def __len__(self): 
   ...:         return 0                                                      
In [3]: m1 = StrictMock(A)                                                    
In [4]: m2 = StrictMock(A) 
```

[bool()](https://docs.python.org/3/reference/datamodel.html#object.__bool__), from its docs, looks for  `__bool__`, which is not defined, then it looks for `__len__`, which is defined in the template class, thus it should be invoked. Given the default behavior of StrictMock, it should raise  `UndefinedBehavior` on `__len__` access, `bool(m1)` should raise, but it does not:

```python
In [5]: bool(m1)                                                              
Out[5]: True
In [6]: bool(m2)                                                              
Out[6]: True
```

Instead, it returns true, as the default value for `bool` (as if both `__bool__` and `__len__` are not defined). If we define a behavior for `__len__`, things works as expected:

```python
In [7]: m1.__len__ = lambda: 0                                                                                                                                
In [8]: bool(m1)                                                                                                                                              
Out[8]: False
```

However, it breaks other instances:

```python
In [9]: bool(m2)                                                                                                                                              
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-a2b1c204a39c> in <module>
----> 1 bool(m2)
~/.pyenv/versions/3.6.6/lib/python3.6/site-packages/testslide/strict_mock.py in __get__(self, instance, _owner)
    119         else:
    120             raise AttributeError(
--> 121                 "{}:\n  Object has no attribute '{}'".format(repr(instance), self.name)
    122             )
    123 
AttributeError: <StrictMock 0x7F27960CD4E0 template=__main__.A - <ipython-input-4-9a7b27e5dca0>:1>:
  Object has no attribute '__len__'
```

This seems to happen because the way `bool` is [imlpemented](https://github.com/python/cpython/blob/3.7/Objects/object.c#L1411-L1433). When we do `m1.__len__ = lambda: 0`, under the hood, StrictMock puts [_DescriptorProxy](https://github.com/facebookincubator/TestSlide/blob/master/testslide/strict_mock.py#L111) at `StrictMock` dictionary. `_DescriptorProxy` correctly implements [`__get__`](https://docs.python.org/3/reference/datamodel.html#object.__get__) to give the correct value for each instance, but `bool()` does not follow the descriptor protocol, and simply checks if the class dictionary has `__len__`. If it happens to implement `__get__` (as `_DescriptorProxy` does), and the instance in particular does not have the attribute, it simply raises the `AttributeError` from `_DescriptorProxy`.

From [`__bool__()`](https://docs.python.org/3/reference/datamodel.html#object.__bool__) docs, I'd expect for it do do something like this:

```python
def bool(obj):
  if hasattr(obj, "__bool__"):
    return getattr(obj, "__bool__")()
  if hasattr(obj, "__len__"):
    return not (return getattr(obj, "__len__")() == 0)
  return True
```

Thus, if a particular instance does not have `__len__` (eg: `__get__` raises `AttributeError`), it sholud work. However, it seems it is doing this:

```python
def bool(obj):
  if "__bool__" in type(obj).__dict__:
    return getattr(obj, "__bool__")()
  if "__bool__" in type(obj).__dict__:
    return not (return getattr(obj, "__len__")() == 0)
  return True
```

The workaround at StrictMock's end (to not depend on any upstream patch), can be:

- Override `StrictMock.__new__` to dynamically create a subclass of StrictMock for each instance.
- Populate this subclass magic methods (`__len__`, `__bool__` etc) accordingly to the template given.

This should make magic methods per `StrictMock` instance, thus covering the `bool()` case, and any other similar scenario. This is a similar solution to what [mock_constructor() does](https://testslide.readthedocs.io/en/1.4.10/mock_constructor/index.html#implementation-details) to go workaround a similar `__new__` bug in Python 3.x.

fornellas:
Refactor StrictMock to address issues with magic methods (see #23).

The mechanic after this diff is:

- Create a subclass of StrictMock.
- Populate this subclass with all existing magic methods from the template class.
- Setting any attributes will set them at the subclass.
- There's validation for valid and invalid magic methods (previously missing).

Closes #23 .

### DIFF
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -3,11 +3,11 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-import sys
 import dis
 import copy
 import functools
 import inspect
+import os.path
 
 from unittest.mock import _must_skip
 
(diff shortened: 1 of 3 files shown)


testslide/strict_mock.py
```diff
@@ -153,41 +154,202 @@ class StrictMock(object):
...
+        "__subclasscheck__",
+        "__truediv__",
+        "__trunc__",
+        "__xor__",
+        # TODO
+        # "__await__",
+        # "__aiter__",
+        # "__anext__",
+        # "__aenter__",
+        # "__aexit__",
```
pallotron: I suppose this is the other PRs for async?

testslide/strict_mock.py
```diff
@@ -153,41 +154,202 @@ class StrictMock(object):
...
+    ]
+
+    _UNSETTABLE_MAGICS = [
```
pallotron: should probably put down a few works as to way they are unsettable.

testslide/strict_mock.py
fornellas: Yes, next PRs are dealing with that.

testslide/strict_mock.py
fornellas: o7

fornellas:
There's some logic to some of the design decisions, I'll add comments to them all so we don't loose the rational.

fornellas:
I just merged this to master by mistake with #68, I'll do another PR on top to address the requests.

PR #75, opened 2019-11-08
Fail async test if coroutine not awaited

fornellas:
What the title says. A test that generates the failure scenario was added.


fornellas:
Only supported by 3.7+...

PR #76, opened 2019-11-08
Add Python 3.8

fornellas:
What the title says.


fornellas:
@tfg13 this is easy as well :-D

PR #95, opened 2019-11-19
Fix callable runtime_attr

fornellas:
Fix "self" not being swallowed when mocking callable runtime attributes.


testslide/strict_mock.py
```diff
@@ -562,6 +562,11 @@ def __setattr__(self, name, value):
                         mock_value = _MethodProxy(validate_awaitable_return, value)
                     else:
                         mock_value = _MethodProxy(value_with_sig_val, value)
+            else:
+                if callable(value):
+                    # We don't really need the proxy here, but it server the
```
xteodora: it serves** typo

PR #99, opened 2019-12-09
Add a flag to disable signature validation

shish:
For times when the signature-validating proxy object causes issues

In particular, one can see this with Thrift objects:

- Thrift-py3 objects are callable, because they are immutable, and use the `__call__` interface to return a modified version of themselves
- Thrift-py3 objects will call `type()` to validate the types of their attributes, and we can't fake our `type()`

So when one does:

```
p1 = Parent1(child=Child())
p2 = Parent2(child=p1.child)
```

if `Parent1` has been replaced with a `StrictMock()`, then `p1.child` will have been replaced with a `_MethodProxy(Child())`, and in `Parent2`'s constructor, `assert type(child) == Child` will fail.

### DIFF
--- a/tests/strict_mock_testslide.py
+++ b/tests/strict_mock_testslide.py
@@ -130,11 +130,6 @@ class CallableObject(object):
         pass
 
 
-class ObjectWithCallableAttribute(object):
-    def __init__(self):
-        self.attr = CallableObject()
-
-
 @context("StrictMock")  # noqa: C901
 def strict_mock(context):
     @context.function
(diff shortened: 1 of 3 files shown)


tests/strict_mock_testslide.py
```diff
@@ -872,3 +872,33 @@ def default_context_manager_True(context):
...
+        self.assertNotEqual(type(mock_obj.attr), type(real_obj.attr))
+
+    @context.example
+    def callable_attribute_not_replaced_with_proxy(self):
+        # If we want type() to give correct results, then we need
+        # to disable method proxies
+        real_obj = self.obj_with_callable_attribute
+        mock_obj = StrictMock(signature_validation=False)
+        mock_obj.attr = real_obj.attr
+        self.assertEqual(type(mock_obj.attr), type(real_obj.attr))
```
fornellas: Please create a dedicated context for this so it does not (potentially) clash with the parent. Eg:

```python
    @context.sub_context
    def disabling_signature_validation(context):
      @context.example
      ...
```

tests/strict_mock_testslide.py
```diff
@@ -872,3 +872,33 @@ def default_context_manager_True(context):
...
+
+    @context.memoize
+    def obj_with_callable_attribute(self):
+        class CallableObject(object):
+            def __call__(self):
+                pass
+
+        class ObjWithCallableAttribute(object):
+            def __init__(self):
+                self.attr = CallableObject()
```
fornellas: Move these to the top of the file with other classes plz.

PR #102, opened 2019-12-09
Add SlowCallback error type

vmagro:
TestSlide previously raised a RuntimeError when it detected an asyncio
slow callback warning message, change that to a custom SlowCallback
exception type that makes the failure case easier to track down, since
the default message is fairly opaque if you haven't seen it before.

This could potentially be extended with more information to help track
down the offending function, but Python does not provide much
information, so in this PR I am just adding the exception class to at
least point in the right direction.

### DIFF
--- a/tests/dsl_unittest.py
+++ b/tests/dsl_unittest.py
@@ -7,10 +7,11 @@ import sys
 import asyncio
 import unittest
 import time
+import re
 
 from unittest.mock import Mock, call, patch
 
-from testslide import Context, AggregatedExceptions, reset
+from testslide import Context, AggregatedExceptions, SlowCallback, reset
 from testslide.dsl import context, xcontext, fcontext
 import os
 import subprocess
@@ -1859,7 +1860,12 @@ class SmokeTestAsync(TestDSLBase):
                 async def example(self):
                     time.sleep(0.1)
 
-            with self.assertRaisesRegex(RuntimeError, "^Executing .+ took .+ seconds$"):
+            with self.assertRaisesRegex(
+                SlowCallback,
+                re.compile(
+                    r"^Executing .+ took .+ seconds\nSlow callback detected.*$", re.M
+                ),
+            ):
                 self.run_first_context_first_example()
 
 
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -207,6 +207,12 @@ class UnexpectedSuccess(Exception):
     """
 
 
+class SlowCallback(Exception):
+    """
+    Raised by TestSlide when an asyncio slow callback warning is detected
+    """
+
+
 class _ExampleRunner:
     def __init__(self, example):
         self.example = example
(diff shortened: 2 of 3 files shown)


testslide/__init__.py
```diff
@@ -207,6 +207,12 @@ class UnexpectedSuccess(Exception):
     """
 
 
+class SlowCallback(Exception):
```
fornellas: Docs refer to this as "Even Loop Lock Up", perhaps we could keep the same naming in both places.

tests/dsl_unittest.py
```diff
@@ -1859,7 +1859,7 @@ def top(context):
                 async def example(self):
                     time.sleep(0.1)
 
-            with self.assertRaisesRegex(RuntimeError, "^Executing .+ took .+ seconds$"):
+            with self.assertRaisesRegex(SlowCallback, "^Executing .+ took .+ seconds$"):
```
fornellas: Tip: if you use multiple line messages here, eg:

```
Event loop got blocked for more than Xms.
Executing .... took ... seconds
```

The test runner properly cut the first line, and only show the long message at the end of the report.

fornellas: PS: It just occurred to me: the exception message could mention `asyncio.get_running_loop().slow_callback_duration = seconds` to inform users how to customize it if needed, just like the docs do.

vmagro: Not sure I understand what you mean here, are you suggesting adding "Event loop got blocked for more than ..." to the `SlowCallback` (`EventLoopLockUp` per your rename suggestion) error message?

testslide/__init__.py
vmagro: I don't see where in the docs it's referenced as "event loop lock up", but I see
> Callbacks taking longer than 100ms are logged. The loop.slow_callback_duration attribute can be used to set the minimum execution duration in seconds that is considered “slow”

In the debugging section of https://docs.python.org/3/library/asyncio-dev.html

fornellas: Oh, I'm sorry, I meant [TestSlide's](https://github.com/facebookincubator/TestSlide/blob/master/docs/testslide_dsl/async_support/index.rst#even-loop-lock-up) documentation which I added yesterday. I agree with you that using Python's naming is better, so please feel fere to amend this at TestSlide's docs (also check the example output, the exception / message needs updating as well).

tests/dsl_unittest.py
fornellas: What I (miserably) tried to mean (sorry!) was that the error message should be as informative and educative as possible. I found myself in the past few days explaining several times why event loop lock ups are something important to surface, let's make this a job of the error message. I also got asked "what if I'm OK with 0.2s (not 0.1s)?" a few times, so, informing about setting `slow_callback_duration` (eg: using a before hook) would also be helpful.

PR #104, opened 2019-12-10
Amend small bits of doc

fornellas:
Documentation is provided here https://testslide.readthedocs.io/, let's not use links to GitHub sources directly.

I don't think there's an of the shelf way to get the current version so we can generate a link to https://testslide.readthedocs.io/en/2.0.1/testslide_dsl/async_support/index.html, so I added a short version of the doc in the exception message.


fornellas:
Meh, nevermind, thanks for the quick review!

PR #106, opened 2019-12-16
Fix StrictMock with signature_validation=False

### LINKED ISSUE #105: disabling_signature_validation=True breaks callable arguments
This:

```python
import asyncio
import testslide

class AsyncContextManager:
  async def __aenter__(self):
    return self

  async def __aexit__(self, *args):
    pass

mock = testslide.StrictMock(
  template=AsyncContextManager,
  default_context_manager=True,
  signature_validation=False,
)

async def test():
  async with mock as cm:
    pass

asyncio.run(test(), debug=True)
```

Fails:

```
TypeError: aenter() takes 0 positional arguments but 1 was given
```
We need to:

- Refactor the tests, so we get all test cases to run with `disabling_signature_validation=True` as well (let's see what else maybe broken).
- Fix what needs fixing.

fornellas:
What the title says. Also:

- Refactor StrictMock tests, removing some clutter.
- Added extra test cases to cover `signature_validation=False` making `type()` work for all use cases.

Closes #105 .

### DIFF
--- a/tests/strict_mock_testslide.py
+++ b/tests/strict_mock_testslide.py
@@ -164,6 +164,15 @@ def strict_mock(context):
                 ),
             )
 
+    @context.memoize
+    def caller_filename(self):
+        current_module = sys.modules[__name__]
+        filename = inspect.getsourcefile(current_module) or inspect.getfile(
+            current_module
+        )
+
+        return filename
+
     @context.shared_context
     def can_access_attributes(context):
         @context.example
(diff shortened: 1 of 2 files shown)


shish:
I feel like I need gifs to express my horror at how many layers of magic we need to deal with to make this work...

fornellas:
"Magic" is a a polite word... let's enjoy the "magic"-storm that it is sometimes....

Thanks for the assist on the design here @shish !!

PR #115, opened 2020-02-27
measure test execution time from process start, display import time

deathowl:


testslide/runner.py
```diff
@@ -270,7 +265,8 @@ def finish(self, not_executed_examples):
                 )
         print("")
         self.print_white(
-            "Finished %s example(s) in %.1fs:" % (total, self.duration_secs)
+            "Finished %s example(s) in %.1fs: . Imports took: %.1fs"
+            % (total, self.duration_secs, self.import_secs)
         )
```
fornellas: Do you think printing import time, even when it is fast, is worth it? Perhaps we can do `if self.import_secs > 2` then also print import times, otherwise only print the total time.

fornellas: LGTM and all tests passed, feel free to merge it :-D

Issue #98, opened 2019-12-02
Fail `mock_callable()` / `mock_async_callable()` with private methods


fornellas:
Generically speaking, mocking private interfaces is bad it breaks the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle): if the private interface of the class changes, tests break. Using dependency injection or the registry pattern are ways around this.

Let's help users write better tests, and fail `mock_callable` / `mock_async_callable` if the method being mocked is private (eg: `_some_method`), by raising `ValueError`. The error message should educate users on why this is a bad pattern, how to go around it, and if it is a very special snow flake, they can still do it by adding the parameter `allow_private=True`.

PR #121, opened 2020-03-02
Disallow patching private interfaces #98

deathowl:


tests/mock_callable_testslide.py
```diff
@@ -860,7 +879,9 @@ def can_not_mock(self):
                 mock_callable(self.target_arg, self.callable_arg)
 
     @context.shared_context
-    def async_methods_examples(context, not_in_class_instance_method=False):
+    def async_methods_examples(
+        context, not_in_class_instance_method=False, alllow_private=True
```
fornellas: Do we need this?

deathowl: Not. stayed there by accident, removing it rn.

PR #123, opened 2020-03-03
Fix TestCase integration

fornellas:
Fix a bug introduced by #119:

```
ValueError: Function must receive parameter #1 named 'context', but given function has parameters: ['test_case_context'].
```

We had no integration tests to cover this, this PR adds them, and fixes the bug.


fornellas:
Also fix black formatting of `testslide/lib.py`.

javierhonduco:
LGTM 🚢

PR #124, opened 2020-03-03
add typechecking for patched functions args

deathowl:


tests/mock_callable_testslide.py
```diff
@@ -164,6 +173,13 @@ def patching_private_functions_with_allow_private(self):
         ).and_assert_called_once()
         t._privatefun()
 
+    @context.example
+    def calling_patched_functions_with_bad_types_raises_TypeError(self):
+        t = TargetStr()
+        self.mock_callable(t, "typedfun").to_return_value("This is patched")
+        with self.assertRaises(TypeError):
```
fornellas: Better use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex here, and assert that the error str matches what we expect. Without it, if there's a TypeError for other reasons (eg: giving more arguments than expected), test will pass, but it won't be actually testing type checking.

testslide/strict_mock.py
```diff
@@ -10,6 +10,7 @@
 import os.path
 
 from unittest.mock import _must_skip
+from .lib import _validate_function_signature
 
 
 def _add_signature_validation(value, template, attr_name):
```
fornellas: s/_add_signature_validation/_wrap_signature_and_type_validation/g

Fair?

fornellas:
This is an **amazing** feature, I feel like it deserves being added to the documentation here 
https://testslide.readthedocs.io/en/2.2.1/patching/mock_callable/index.html#signature-validation

I'd add another section "Type Checking Validation" and give an example, similar to the one for signatures.

It is amazing how small this PR is, and how powerful it'll be :-D

PR #129, opened 2020-03-24
[strict_mock] Speed up first invocation

david-caro:
This is done by getting only the minimum needed amount of frames from
the stack, and avoid retrieving the file contents for each when trying
to get the caller, when initializing the StrictMock object.

This avoids both, unnecessary file searches, and unnecessary file reads,
speeding up the first invocation from ~0.18s to ~0.00015 (only did a
couple tests, but the speed is up considerably).

Before:
```
test nothing
time for _get_caller: 0.17940688133239746
  make sure the first instance of strictmock does not timeout any tests: SlowCallback: ... took 0.180 seconds
```

After:
```
dummy test
time for _get_caller: 0.00014853477478027344
  make sure the first instance of strictmock does not timeout any tests
```

### DIFF
--- a/tests/strict_mock_testslide.py
+++ b/tests/strict_mock_testslide.py
@@ -341,6 +341,19 @@ def strict_mock(context):
                     getattr(self.strict_mock, attr_name)
 
             @context.example
+            def shows_the_correct_file_and_linenum_when_raising_when_an_undefined_attribute_is_accessed(
+                self,
+            ):
+                attr_name = "non_callable"
+                with self.assertRaisesWithRegexMessage(
+                    UndefinedAttribute,
+                    f"'{attr_name}' is not set.\n"
+                    f"{self.strict_mock_rgx} must have a value set "
+                    "for this attribute if it is going to be accessed.",
+                ):
+                    getattr(self.strict_mock, attr_name)
+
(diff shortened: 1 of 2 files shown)


fornellas:
@david-caro thanks for this improvement! I'm not totally sure there's test coverage for this exact functionality, do you mind doing a simple (even manual) test on this to confirm it still shows up the same file name:line please?

LGTM for merging if working :-D

david-caro:
Tests found an issue, awesome :)

david-caro:
I'm doubting if I should add a new test that it's exactly the same as the one that failed, but with the name `shows_the_correct_file_and_linenum_when_raising_when_an_undefined_attribute_is_accessed`, so there's a specific test that will not be changed to test the file:linenum, wdyt?

fornellas:
I love the idea of having tests focused on "business logic", so yes, I agree that having a 5 line test for this functionality is awesome. This means we can relax other tests on path:lineno, so signal is more fine grained :-D

I'm curious though why Travis CI is not triggering tests for this PR... anyway, just run `make` and you get (mostly) the same signal.

fornellas:
I added you to the project, feel free to merge once tests pass :-D

PR #131, opened 2020-03-25
Add AnyWithCall matcher

oxo42:
This creates a completely generic `Matcher` where you can pass a
function in to be called.  Something like

```
self.mock_callable(os, 'remove')\
  .for_call(AnyWithCall(lambda path: path.endswith("py"))\
  .to_return_value(None)\
  .and_assert_called_once()
```

### DIFF
--- a/docs/patching/argument_matchers/index.rst
+++ b/docs/patching/argument_matchers/index.rst
@@ -111,3 +111,11 @@ Generic
 	"``AnyTruthy()``", "Any object where ``bool(obj) == True``"
 	"``AnyFalsey()``", "Any object where ``bool(obj) == False``"
 	"``AnyInstanceOf()``", "Any object where ``isinstance(obj) == True``"
+	"``AnyWithCall(call)``", "Any object where ``call(obj) == True``"
+
+.. code-block:: python
+
+  self.mock_callable(os, 'remove')\
+    .for_call(AnyWithCall(lambda path: path.endswith("py"))\
+    .to_return_value(None)\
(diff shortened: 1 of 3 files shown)


david-caro:
This looks good to me, but I'll let somebody else review too (being my first xd).

docs/patching/argument_matchers/index.rst
```diff
@@ -110,4 +110,12 @@ Generic
...
\ No newline at end of file
+	"``AnyInstanceOf()``", "Any object where ``isinstance(obj) == True``"
+	"``AnyWithCall(call)``", "Any object where ``call(obj) == True``"
+
+.. code-block:: python
+
+  self.mock_callable(os, 'remove')\
+    .for_call(AnyWithCall(lambda path: path.endswith("py"))\
+    .to_return_value(None)\
+    .and_assert_called_once()
```
fornellas: I'm not sure what to say about having this example floating here. We had at some point multiple examples, but doc became cumbersome, so I trimmed to let only the description text. To my eyes at least, it is good enough above, but I'm biased to take any sides here...

oxo42: The reason I wanted an example is just looking at the class signature, it is not obvious how this works.  I am inclined to leave this in and in the future if you clean it out, I won't complain.

david-caro:
@oxo42 why did you close this? I still think it's useful

fornellas:
Definitely, this should go to master!

PR #132, opened 2020-03-31
[strict_mock] Support mocks in signature validation

### LINKED ISSUE #130: Type checker fails with mocks
[This PR](https://github.com/facebookincubator/TestSlide/pull/124) added type checking for function calls. It works as expected, but it is failing when a call is made using a mock, but it should not.

Suggestion on how to fix this:

- Have a constant with a list of known mock classes: Python's unittest (eg: the ones [listed here](https://testslide.readthedocs.io/en/2.3.0/strict_mock/index.html#yet-another-mock)) and StrictMock.
- Have a map of mock class to a function that extracts its template:
  - For `StrictMock`, get the given `template` (IIRC mock_callable already does this, there's a trick there).
  - For Python's unittest, get the [`spec`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock).

Using that:

- If the given argument is a mock:
  - Without a template / spec, then do no type checking.
  - With a template, do type checking using the template.

david-caro:
We were not propely checking the signature validation when the
parameters passed were mocks and failing with TypeError.

This tries to extract the spec or template of the parameter to do the
type checking allowing to pass mocks to the functions without failing
the signature type validation.

NOTE: this assumes that all unittest.mock mocks inherit from
NonCallableMock, that assumption might change with time, though
currently seems to be the case:
```
$ grep '^class .*Mock(' /usr/lib64/python3.7/unittest/mock.py
class NonCallableMock(Base):
class Mock(CallableMixin, NonCallableMock):
class NonCallableMagicMock(MagicMixin, NonCallableMock):
class MagicMock(MagicMixin, Mock):
class PropertyMock(Mock):
```

Fixes #130

### DIFF
--- a/tests/strict_mock_testslide.py
+++ b/tests/strict_mock_testslide.py
@@ -511,38 +511,52 @@ def strict_mock(context):
                                         self.assertEqual(method("hello"), "mock: hello")
 
                                     @context.example
-                                    def works_with_Mock_parameter(self):
-                                        class Dummy_param:
+                                    def works_when_the_parameter_passed_is_an_instance_of_Mock(
+                                        self,
+                                    ):
+                                        class DummyParam:
                                             def get_string():
-                                                return "Dummy_param: not mocked"
+                                                return "DummyParam: not mocked"
 
                                         class Dummy:
-                                            def method(self, param: Dummy_param):
+                                            def method(self, param: DummyParam):
                                                 return f"Dummy: {param.get_string()}"
 
                                         my_strict_mock = StrictMock(template=Dummy)
-                                        my_mock = Mock(spec=Dummy_param)
+                                        my_mock = Mock(spec=DummyParam)
                                         my_mock.get_string.return_value = "mock_param"
 
-                                        my_strict_mock.method = lambda param: f"mock_method: {param.get_string()}"
-                                        self.assertEqual(my_strict_mock.method(param=my_mock), "mock_method: mock_param")
+                                        my_strict_mock.method = (
+                                            lambda param: f"mock_method: {param.get_string()}"
+                                        )
+                                        self.assertEqual(
+                                            my_strict_mock.method(param=my_mock),
+                                            "mock_method: mock_param",
+                                        )
 
                                     @context.example
-                                    def works_with_StrictMock_parameter(self):
-                                        class Dummy_param:
+                                    def works_when_the_parameter_passed_is_an_instance_of_StrictMock(
+                                        self,
+                                    ):
+                                        class DummyParam:
                                             def get_string():
-                                                return "Dummy_param: not mocked"
+                                                return "DummyParam: not mocked"
 
                                         class Dummy:
-                                            def method(self, param: Dummy_param):
+                                            def method(self, param: DummyParam):
                                                 return f"Dummy: {param.get_string()}"
 
                                         my_strict_mock = StrictMock(template=Dummy)
-                                        my_mock = StrictMock(template=Dummy_param)
+                                        my_mock = StrictMock(template=DummyParam)
                                         my_mock.get_string = lambda: "mock_param"
 
-                                        my_strict_mock.method = lambda param: f"mock_method: {param.get_string()}"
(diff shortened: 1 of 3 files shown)


tests/strict_mock_testslide.py
```diff
@@ -509,6 +510,40 @@ def works_with_wraps(self):
...
+                                    def works_with_Mock_parameter(self):
+                                        class Dummy_param:
+                                            def get_string():
+                                                return "Dummy_param: not mocked"
+
+                                        class Dummy:
+                                            def method(self, param: Dummy_param):
+                                                return f"Dummy: {param.get_string()}"
+
+                                        my_strict_mock = StrictMock(template=Dummy)
```
fornellas: Shouldn't this be:
```
my_strict_mock = Mock(spec=Dummy)
```
as per test title?

tests/strict_mock_testslide.py
```diff
@@ -509,6 +510,40 @@ def works_with_wraps(self):
...
+                                        my_mock = StrictMock(template=Dummy_param)
+                                        my_mock.get_string = lambda: "mock_param"
+
+                                        my_strict_mock.method = lambda param: f"mock_method: {param.get_string()}"
+                                        self.assertEqual(my_strict_mock.method(param=my_mock), "mock_method: mock_param")
+
```
fornellas: I think we could have tests for "generic" mocks without spec / template: `Mock()` / `StrictMock()` as well, fair?

testslide/lib.py
```diff
@@ -20,26 +17,3 @@ def _bail_if_private(candidate: str, allow_private: False):
             "Please consider using patterns like dependency injection instead. "
             "If you really need to do this use the allow_private=True argument."
         )
-
-
-def _validate_function_signature(argspec: inspect.FullArgSpec, args, kwargs):
```
fornellas: IIRC, the reason for this to live here, was because we should also in the future, use it for `mock_callable()` as well.

fornellas:
I think we could use a "generic" interface to define which are the mock objects, and a decoupled & extensible way to extend things, see #130 description for a (simple) suggestion on how to do it. To keep your good work, perhaps `_unwrap_mock` could just lookup a dictionary for the class / function. We add this to the docs, so when integrating with any framework, it is trivially extensible.

tests/strict_mock_testslide.py
david-caro: What I'm validating is that a strict mock, that has to validate a call to a mocked method, validates correctly when one of the params passed to that call is actually a mock with a spec

testslide/lib.py
david-caro: I had to move it because I'm using the class `StrictMock` to match the argument against it, so that would have introduced a cyclic import.

I could try to pass to the function the list of mock classes to check or similar, but seemed simpler to just move than create more layers.

david-caro:
> I think we could use a "generic" interface to define which are the mock objects, and a decoupled & extensible way to extend things, see #130 description for a (simple) suggestion on how to do it. To keep your good work, perhaps _unwrap_mock could just lookup a dictionary for the class / function. We add this to the docs, so when integrating with any framework, it is trivially extensible.

That's what I did first, but then realized that any mock in unittest is actually a subclass of `NonCallableMock`, so I used that :)
I did miss that other frameworks would have different mock classes though, I'll move back to the dict kind of solution.

testslide/lib.py
deathowl: @fornellas yes, that was the exact reason.

david-caro: I'll try to add the "mock_extractors" param or similar with the list of mock classes and how to get the spec/template out of them then, though to be fair, `mock_callable` already depends on `strict_mock`.

fornellas:
I just noticed how the tests (including existing ones) are not covering all cases: sync/async instance/static/class/magic methods for signature validation. Historically, corner cases among them were the source of many bugs (and headaches).

Improving this is kinda orthogonal to the goal here, so let's merge this, and I'l cut a PR to tighten these tests on top.

Thanks @david-caro !

Issue #135, opened 2020-04-02
Add options to disable type checking


fornellas:
Currently `StrictMock` reuses the signature validation flag `signature_validation` to enable/disable type checking and `mock_callable()`, `mock_async_callable()` and `mock_constructor()` have no option to disable type checknig.

Let's add a `type_validation=True` option to allow users to disable type checking:

- `StrictMock`: at `__new__` & `__initt__` plus a test to cover this.
- `mock_callable()`: add option plus a test to cover this.
- `mock_async_callable()` and `mock_constructor()`: as they re-use mock_callable engine, make sure they're passing the option along.

fornellas:
Tip: #134 is improving the tests. On top of it, it should be trivial to add a few lines of test for each case that covers all scenarios (module, instance/class/static methods) for all cases, just drop me a message and I can pin point where they are.

PR #138, opened 2020-04-06
#135 Add ability to disable typechecking

deathowl:


fornellas:
I'm merging this now, and immediately after #140, which fixes the pending issues here.

Nice collaborating with @deathowl on this :-D

Issue #136, opened 2020-04-02
Type check callable returns


fornellas:
We currently only type check callable arguments from `mock_callable()`, `mock_async_callable()`, `mock_constructor()` and `StrictMock`. Let's also check the return value of the call if type checking is available.

fornellas:
Closed by #139 & #144.

PR #139, opened 2020-04-06
Type check return value for mock_callable and mock_async callable

deathowl:
Fixing #136

### DIFF
--- a/tests/mock_callable_testslide.py
+++ b/tests/mock_callable_testslide.py
@@ -30,7 +30,7 @@ class TargetStr(object):
         a: str,
         b: typing.Iterable[typing.Union[str, float]],
         c: typing.Optional[str],
-    ):
+    ) -> str:
         return "asd"
 
     def _privatefun(self):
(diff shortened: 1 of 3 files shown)


tests/mock_callable_testslide.py
```diff
@@ -179,6 +179,18 @@ def patching_private_functions_with_allow_private(self):
         ).and_assert_called_once()
         t._privatefun()
 
+    @context.example
+    def patching_typed_functions_raises_typeerror_with_invalid_responsetype(self):
+        t = TargetStr()
+        with self.assertRaises(TypeError):
```
fornellas, line 185: How is this error message looking like? I'm wondering if it needs any tweaking, or if it is good enough as it comes from typeguard.

testslide/mock_callable.py
```diff
@@ -713,6 +713,7 @@ def to_return_value(self, value):
         """
         Always return given value.
         """
+        _validate_return_type(self._original_callable, value)
```
fornellas, line 716: This is not enough: other stuff such as `.to_return_values()` won't work. I believe that if we add it to `_CallableMock.__call__`, it should cover 100% of cases.

I'll cut a PR on top, adding tests specifically for the changes here, covering all cases.

PR #147, opened 2020-04-08
Implement type checking for StrictMock non-callable attributes

### LINKED ISSUE #137: Add type checking for non callable attributes at StrictMock
We currently only check type for callable attributes for `StrictMock` instances.

Let's add type checking when setting non-callable attributes.

fornellas:
Ditto.

Closes #137 .

### DIFF
--- a/testslide/lib.py
+++ b/testslide/lib.py
@@ -42,18 +42,12 @@ def _is_a_mock(maybe_mock: Any) -> bool:
     )
 
 
-def _validate_argument_type(
-    annotations: Dict[str, Any], argname: str, value: Any
-) -> None:
-    type_information = annotations.get(argname)
-    if not type_information:
-        return
-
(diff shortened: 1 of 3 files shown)


testslide/lib.py
```diff
@@ -74,12 +68,18 @@ def _validate_function_signature(
             else:
                 argname = argspec.args[idx]
             try:
-                _validate_argument_type(argspec.annotations, argname, args[idx])
+                expected_type = argspec.annotations.get(argname)
+                if not expected_type:
+                    continue
```
david-caro, line 73: Why move this previous lines out of the `_validate_argument_type` function? it looks like we have them everywhere we use that function no?

fornellas: Because from StrictMock, we don't have argspec, only the type available. By doing this change here, we can use `_validate_argument_type` for function arguments and and class attributes type validation :-D

PR #148, opened 2020-04-08
Add Long Formatter

fornellas:
Add `--format long` option to have this one-liner view of test results:

image

### DIFF
--- a/tests/cli_unittest.py
+++ b/tests/cli_unittest.py
@@ -254,7 +254,7 @@ class TestCliQuiet(TestCliBase):
         )
 
 
-class TestCliDocumentation(TestCliBase):
+class TestCliDocumentationFormatter(TestCliBase):
     def setUp(self):
         super().setUp()
         self.argv = ["--format", "documentation"] + self.argv
@@ -520,7 +520,7 @@ class TestCliDocumentation(TestCliBase):
         )
 
 
-class TestCliProgress(TestCliBase):
+class TestCliProgressFormatter(TestCliBase):
     def setUp(self):
         super().setUp()
         self.argv.append("--format")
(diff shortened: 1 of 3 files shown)


tests/cli_unittest.py
```diff
@@ -154,110 +154,404 @@ def cyan(text):
         return "\x1b[0m\x1b[36m{}\x1b[0m".format(text)
 
 
-class TestCliList(TestCliBase):
+# class TestCliList(TestCliBase):
```
david-caro, line 157: Why is this commented out?

fornellas: BUG! Nice catch!

tests/cli_unittest.py
```diff
@@ -546,3 +546,112 @@ def test_ouputs_colored_dots_with_terminal(self):
...
+                + self.red("test_failing: ")
+                + "AssertionError: "
+                + "\r\n"
+                + self.white("tests.sample_tests.SampleTestCase: ")
+                + self.green("test_passing")
+                + "\r\n"
+                + self.white("tests.sample_tests.SampleTestCase: ")
+                + self.yellow("test_skipped")
+                + "\r\n"
+                # TODO Rest of the output
```
fabriziocucci, line 593: Is this really a `TODO` or just `...rest of the output`?

Same for the other tests...

testslide/runner.py
```diff
@@ -279,6 +279,122 @@ def finish(self, not_executed_examples):
...
+                focus="*" if example.focus else "",
+                example=example,
+                pass_text="" if self._color_output() else ": PASS",
+            )
+        )
+
+    def fail(self, example, exception):
+        if isinstance(exception, AggregatedExceptions) and 1 == len(
+            exception.exceptions
+        ):
```
fabriziocucci, line 307: Ignorance of the codebase (my bad): is there a chance that `... and len(exception.exceptions) > 1` ?

tests/cli_unittest.py
fornellas: It is a TODO. The output at this point contains timing information, which would require using regex, which's a pain with colored output. It is currently a compromise: covering 90% of features with 10% of effort.

fabriziocucci: Thanks for the context. Maybe we could add this info to the comment to make it more clear?

PR #149, opened 2020-04-08
Take into account varargs in function signature validation

### LINKED ISSUE #145: Fix argument type validation for *args
This:

```
testslide.lib._validate_function_signature(
  object.__new__, (1, 2, 3), {"four": 5, "six": 6}
)
```
should pass, but is failing with:
```
IndexError: list index out of range
File "testslide/lib.py", line 78, in _validate_function_signature
  argname = list(signature.parameters.keys())[idx]
```
This is blocking #141 .

david-caro:
Fixes #145

### DIFF
--- a/tests/lib_testslide.py
+++ b/tests/lib_testslide.py
@@ -36,6 +36,12 @@ def _validate_function_signature(context):
                 arg1="arg1", arg2="arg2", kwarg1="kwarg1", kwarg2="kwarg2"
             )
 
+        @context.example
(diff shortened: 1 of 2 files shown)


fornellas:
Can you please add a test at `tests/lib_testslide.py` for the case from #145 please?

```
testslide.lib._validate_function_signature(
  object.__new__, (1, 2, 3), {"four": 5, "six": 6}
)
```

If this passes (and files are black formatted), it is good to go!

Issue #142, opened 2020-04-07
Update documentation


fornellas:
- `StrictMock`
  - `type_validation` attribute (at `__init__` and `docs/`)
- `mock_callable`, `mock_async_callable`, `mock_constructor`:
  - - `type_validation` attribute.

PR #151, opened 2020-04-09
StrictMock type validation docs

fornellas:
Part of #142 .

I took the opportunity to shuffle things around at the doc, reduce some redundancy, and focus on functionalities. Hopefully this will be a better answer to the recurring question "why another mock?".

@fabriziocucci you're a good person to judge if this is any good as you should have less biases, hit me!

### DIFF
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -476,6 +476,9 @@ class StrictMock(object):
         the attribute will change. Setting this value to False disables
         signature validation, and should only be used when type() is required
         to not change.
+        type_validation: use type hinting information from template to validate
+        that mock attribute types match them. Type validation also happens for
+        callable attributes (instance/static/class methods) calls.
         """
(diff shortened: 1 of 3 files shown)


testslide/strict_mock.py
```diff
@@ -476,6 +476,9 @@ def __init__(
         the attribute will change. Setting this value to False disables
         signature validation, and should only be used when type() is required
         to not change.
+        type_validation: use type hinting information from template to validate
+        that mock attribute types match them. Type validation also happens for
+        callable attributes (instance/static/class methods) calls.
```
david-caro, line 481: Maybe we can use some better formatted docstrings, google docstrings or reSt text eventually:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

fornellas: I'm all in to merging PRs for this.

PR #154, opened 2020-04-09
Add DSL debugging option

### LINKED ISSUE #89: DSL debug
Sometimes it is hard to understand what hooks (before / after / around) are being executed. It would be nice to have a --debug option that would make the test runner show what hook is being executed, something like this perhaps:

```
Some Context
  before: do_something() file.py:33
    Some Example
      example: some_example() file.py:38
      PASS
  after: cleanup() file.py:45
```

fornellas:
Adds `--dsl-debug` that makes it show when each of the DSL elements that the user defined is called:

- example
- before
- after
- around
- memoize
- memoize_before
- function

image

This should facilitate development and debugging on complex scenarios.

Closes #89.

### DIFF
--- a/tests/cli_unittest.py
+++ b/tests/cli_unittest.py
@@ -508,9 +508,9 @@ class TestCliDocumentationFormatter(TestCliBase):
             ),
         )
 
-    def test_empty_trim_strace_path_prefix(self):
+    def test_empty_trim_path_prefix(self):
         """
-        Trims nothing if '' passed to --trim-stack-trace-path-prefix.
+        Trims nothing if '' passed to --trim-path-prefix.
         """
         self.argv.append("--trim-path-prefix")
         self.argv.append("")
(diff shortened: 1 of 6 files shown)


tests/cli_unittest.py
```diff
@@ -512,13 +512,38 @@ def test_empty_trim_strace_path_prefix(self):
         """
         Trims nothing if '' passed to --trim-stack-trace-path-prefix.
```
fabriziocucci, line 513: Should be `--trim-path-prefix`?

testslide/__init__.py
```diff
@@ -217,8 +219,10 @@ class SlowCallback(Exception):
 
 
 class _ExampleRunner:
-    def __init__(self, example):
+    def __init__(self, example, formatter):
         self.example = example
+        self.formatter = formatter
+        self.trim_path_prefix = self.formatter.trim_path_prefix  # FIMXE
```
fabriziocucci, line 225: `FIXME`?

tests/cli_unittest.py
fabriziocucci, line 513: Should be `--trim-path-prefix`?

PR #163, opened 2020-04-17
Fix *args typing validation

fornellas:
We were not correctly handling this case that is contained in the added test.

Along the way:

- I separated logic regarding signature validation & type validation.
- Changed methods to have proper names.
- Got rid of the ugly dependency on unittest's `_must_skip` private method.
  - Decoupled this logic from both signature & type validation.
  - The new method `_skip_first_arg` must be called from inside `_wrap_signature_and_type_validation` because if it is done from other functions that need its return, the patch will already be in place, and it'll not work.

Credits to @deathowl for helping with all the debugging and ground work prior to this PR.

### DIFF
--- a/tests/lib_testslide.py
+++ b/tests/lib_testslide.py
@@ -43,6 +43,15 @@ def _validate_callable_arg_types(context):
                 False, object.__new__, (1, 2, 3), {"four": 5, "six": 6}
             )
 
+        @context.example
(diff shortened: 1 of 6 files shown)


fornellas:
I like how 0.01% coverage change breaks the signal... this is literally less than 2 lines.

PR #164, opened 2020-04-18
[signature validation] Skip TypeVar signatures

david-caro:
In order to be able to check TypeVars we have to be able to check the
whole signature type hints at the same time (as typevars are dynamically
defined).

This just skips them for now to avoid failing the type checks.

It's using the `str` method of the type hint to traverse the type
definition and see if there are any typevars (prefixed with `~`), it's a
hacky way, but quite effective until we have the need to traverse the
type hint ourselves, at that point we can use that instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/*.pyc
 **/__pycache__/
+.*.sw?
 .coverage
 .coverage.*.*.*
 .mypy_cache/
--- a/testslide/lib.py
+++ b/testslide/lib.py
@@ -82,17 +82,35 @@ def _validate_function_signature(
                 expected_type = argspec.annotations.get(argname)
                 if not expected_type:
                     continue
+
+                if "~" in str(expected_type):
+                    # this means that we have a TypeVar type, and those require
+                    # checking all the types of all the params as a whole, but we
+                    # don't have a good way of getting right now
+                    # TODO: #165
+                    continue
+
                 _validate_argument_type(expected_type, argname, args[idx])
             except TypeError as type_error:
                 type_errors.append(f"{repr(argname)}: {type_error}")
+
     for argname, value in kwargs.items():
         try:
             expected_type = argspec.annotations.get(argname)
             if not expected_type:
                 continue
+
+            if "~" in str(expected_type):
+                # this means that we have a TypeVar type, and those require
+                # checking all the types of all the params as a whole, but we
+                # don't have a good way of getting right now
+                # TODO: #165
+                continue
+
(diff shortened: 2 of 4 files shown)


testslide/lib.py
```diff
@@ -82,17 +82,33 @@ def _validate_function_signature(
                 expected_type = argspec.annotations.get(argname)
                 if not expected_type:
                     continue
+
+                if '~' in str(expected_type):
+                    # this means that we have a TypeVar type, and those require
+                    # checking all the tipes of all the prams as a whole, but we
```
fabriziocucci, line 88: Nit: `tipes` and `prams`.

testslide/lib.py
```diff
@@ -82,17 +82,33 @@ def _validate_function_signature(
...
+
     for argname, value in kwargs.items():
         try:
             expected_type = argspec.annotations.get(argname)
             if not expected_type:
                 continue
+
+            if '~' in str(expected_type):
+                # this means that we have a TypeVar type, and those require
+                # checking all the tipes of all the prams as a whole, but we
```
fabriziocucci, line 104: Nit: `tipes` and `prams`.

.gitignore
```diff
@@ -1,5 +1,6 @@
 **/*.pyc
 **/__pycache__/
+.*.sw?
```
fornellas, line 3: vim files, right?

testslide/lib.py
```diff
@@ -82,17 +82,33 @@ def _validate_function_signature(
                 expected_type = argspec.annotations.get(argname)
                 if not expected_type:
                     continue
+
+                if '~' in str(expected_type):
```
fornellas, line 86: Is there a more thorough way of detecting this? Eg: `isinstance`?

.gitignore
david-caro: Yep :)

testslide/lib.py
david-caro: The problem is that it can be nested, like `Type[~Something]`, `Optional[Tuple[int, Tuple[str, ~Something]]]`, and I did not want to have to implement the unwrapping of that, but as `str` already does and `~` is kinda unique, seems like a good trade-off.

But yes, I think that eventually we will have to unwrap that, at that point we can use that here instead.

testslide/lib.py
david-caro: :+1:

testslide/lib.py
fornellas: That's reasonable for the time being, NP.

PR #166, opened 2020-04-18
Improve Makefile mechanics & output

fornellas:
Improve Makefile:

- Nicer slimmer output.
- Travis CI signal remains full verbose.
- Colored output works on fails.
- Fixed mypy not covering `tests/`.
- Moved `mock_async_callable` tests to its own file (as is).
- Making a test file now works: `make tests/cli_unittest.py`.

And finally, `make -j`:

- Before: 19.6s.
- Now: 9.0s.

# Quiet on success

image

# Show output on failures

image

# Show output with `V=1`

image

### DIFF
--- a/util/run_silent_if_successful.py
+++ b/util/run_silent_if_successful.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+
+import pty, sys, subprocess, os, threading
+
+
+master_pty_fd, slave_pty_fd = pty.openpty()
+read_data = []
+
+
+def master_pty_fd_reader():
+    while True:
+        try:
+            data = os.read(master_pty_fd, 1024)
+        except OSError:
+            return
+        else:
+            if data:
+                read_data.append(data)
+            else:
+                return
+
+
+master_pty_fd_reader_thread = threading.Thread(target=master_pty_fd_reader)
+
+master_pty_fd_reader_thread.start()
+
+completed_process = subprocess.run(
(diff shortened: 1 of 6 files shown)


tests/mock_async_callable_testslide.py
```diff
@@ -0,0 +1,638 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
david-caro, line 1: Can you split this on it's own PR? It looks unrelated (maybe I'm missing something xd)

util/run_silent_if_successful.py
```diff
@@ -0,0 +1,38 @@
...
+)
+
+os.close(slave_pty_fd)
+
+master_pty_fd_reader_thread.join()
+
+if completed_process.returncode:
+    for data in read_data:
+        os.write(sys.stdout.fileno(), data)
+    sys.exit(completed_process.returncode)
```
david-caro, line 38: We might want to save the logs in a file or something eventually, for debugging after the fact.

tests/mock_async_callable_testslide.py
fornellas: o7

util/run_silent_if_successful.py
fornellas: One can do `make V=1` to see all output, but I also see values on having logs.

I'm not sure how we'd implement it: `Makefile` has many different rules, and we'd need a way to reset the log at every execution. Plus, there's concurrency (`make -j`) to consider.

PR #167, opened 2020-04-19
Split mock_async_callable tests

fornellas:
Split `mock_callable` & `mock_async_callable` tests in its own files. Code was moved "as is", with only minor adjustments required for the build to pass.


fabriziocucci:
I wanted to suggest this, really love it!

I walso also thinking about renaming all `*_testslide.py` tests as `testslide_*.py` so that they get all nicely sorted in the `tests` folder (probably creating a `testslide` subfolder would be too much).

PR #168, opened 2020-04-19
Better handling of received signals

fornellas:
Improve signal handling (specially ^C):

- `testslide` does not spit stack trace for ^C.
- `util/run_silent_if_successful.py`:
  - Forward received signals to subprocess and wait for it to terminate.
  - Handle file not found errors gracefully.
- Add `util/` to black, mypy & flake8.

Before:
image
After:
image

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,4 +4,6 @@ parallel = True
 [report]
 precision = 2
 fail_under = 92.31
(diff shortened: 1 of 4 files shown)


fornellas:
Yep! It was missing `util/*` at `.coveragerc`, added!

PR #172, opened 2020-04-20
Fix recursive type validation

fornellas:
typeguard recursively digs into types when possible (eg: `Union`, `Tuple`). This PR patches and wraps `typeguard.check_type` to make it aware of mocks and play nicely with them, as it is expected for TestSlide.

This is a stepping stone only, I'll cut a PR for upstream typeguard, so we can remove the ugly patch from here eventually.

I just shuffled tests to make them flatter, without touching any logic. The tests added here are these:

```
  recursion and mocks
    typing.Union
      passes with StritMock without template
      it works with unittest.mock.Mock without spec
      passes with StritMock with valid template
      passes with unittest.mock.Mock with valid spec
      fails with StritMock with invalid template
      fails with unittest.mock.Mock with invalid spec
    typing.Tuple
      passes with StritMock without template
      it works with unittest.mock.Mock without spec
      passes with StritMock with valid template
      passes with unittest.mock.Mock with valid spec
      fails with StritMock with invalid template
      fails with unittest.mock.Mock with invalid spec
```

Many **thanks** to @fabriziocucci  & @david-caro  to doing all the heavy duty ground work before this fix. It would not have happened without your support!

### DIFF
--- a/testslide/lib.py
+++ b/testslide/lib.py
@@ -60,11 +60,18 @@ def _validate_callable_signature(
 
 
 def _validate_argument_type(expected_type, name: str, value) -> None:
-    if _is_a_mock(value):
-        template = _extract_mock_template(value)
-        if not template:
-            return
-    typeguard.check_type(name, value, expected_type)
+    original_check_type = typeguard.check_type
+
+    def wrapped_check_type(argname, inner_value, inner_expected_type, *args, **kwargs):
+        if _is_a_mock(inner_value):
+            if _extract_mock_template(inner_value) is None:
+                return
+        return original_check_type(
+            argname, inner_value, inner_expected_type, *args, **kwargs
+        )
+
+    with unittest.mock.patch.object(typeguard, "check_type", new=wrapped_check_type):
+        typeguard.check_type(name, value, expected_type)
 
 
 def _validate_callable_arg_types(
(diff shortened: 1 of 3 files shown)


testslide/lib.py
```diff
@@ -60,11 +60,18 @@ def _validate_callable_signature(
...
+    def wrapped_check_type(argname, inner_value, inner_expected_type, *args, **kwargs):
+        if _is_a_mock(inner_value):
+            if _extract_mock_template(inner_value) is None:
+                return
+        return original_check_type(
+            argname, inner_value, inner_expected_type, *args, **kwargs
+        )
+
+    with unittest.mock.patch.object(typeguard, "check_type", new=wrapped_check_type):
+        typeguard.check_type(name, value, expected_type)
```
david-caro, line 74: Just note that any other module that did an import of typeguard, will see the patched version for the little time that this with runs, so it might lead to very weird errors if running in parallel, maybe we should monkey-patch at the beginning only, as un-patching might also interfere, wdyt?

fornellas: I agree, but not sure how to fix. I put the patch with the smallest scope possible (might even take performance away...) precisely because of that... Not sure which's worse.  Hopefully typeguard will get this improved upstream, and we'll get rid of this soon.
Feel free to cut a PR to change this if you feel passionate, I'll gladly merge it.

PR #177, opened 2020-04-21
improved error logging for return type error

### LINKED ISSUE #161: Improve typing error messages
Currently typing error messages only give the stack trace and says "this return/argument is of wrong type". While accurate, this makes it hard to trace back **where** the mock that is implementing this check was created.

Let's improve this message, so it also tells in which file:lineno the `StrictMock`, `mock_callable`, `mock_async_callable` or `mock_constructor` that raised was created.

jottosson1:
Changed the logging for whenever there is a typing error on the return type it will log what file / line number the error happened to make it easier for the user to find the problem.

Closes #161.

### DIFF
--- a/tests/lib_testslide.py
+++ b/tests/lib_testslide.py
@@ -6,6 +6,7 @@
 from typing import Type, TypeVar
 from testslide.dsl import context, xcontext, fcontext, Skip  # noqa: F401
 import testslide.lib
+import testslide.mock_callable
 from . import sample_module
 from testslide import StrictMock
 import unittest.mock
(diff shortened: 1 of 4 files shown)


tests/lib_testslide.py
```diff
@@ -85,6 +85,20 @@ def gives_correct_error_message_for_invalid_types(self):
                 self.skip_first_arg, self.callable_template, (1, 2, 3, 4), {}
             )
 
+    @context.example("Invalid return Type raises TypeError")
+    def assert_raised_typeerror(self):
+        with self.assertRaises(
+            TypeError,
+                msg=(
+                "Call with incorrect return types at "
+                "/Users/jottosson/TestSlide/tests/sample_module.py:101:\n"
```
fornellas, line 94: Please don't use your own paths here, as it'll break on other machines. Check the `trim_path_prefix` parameter that comes from the CLI. You should use it on the code that generates the exception to trim the prefix.

tests/lib_testslide.py
```diff
@@ -85,6 +85,20 @@ def gives_correct_error_message_for_invalid_types(self):
...
+                    "expected <class 'int'> got str"
```
fornellas, line 95: I had in mind something like:

```
f"expected <class 'int'> got str: {repr(received_wrong_value)}"
```

so people can see what object is that.

testslide/mock_callable.py
```diff
@@ -462,7 +462,8 @@ def _get_runner(self, *args, **kwargs):
 
     def _validate_return_type(self, runner, value):
         if runner.TYPE_VALIDATION and runner.original_callable is not None:
-            _validate_return_type(runner.original_callable, value)
+            _validate_return_type(runner, value)
```
fornellas, line 465: If you send runner here,, it'll check value against `_CallableMock.__call__`, which returns `Any`, so I think nothing will be checked.

jottosson1: I dont think so or im missing something, i also changed the _validate_return_type function to take in to account runner is being sent in there

fornellas: It did break a mock_callable test, check Travis CI. Just run `make` on your machine to run all tests before pushing.

PS: signal in Travis CI is hidden because of [this issue](https://github.com/facebookincubator/TestSlide/issues/182).

fabriziocucci:
Closing in favor of #184!

PR #190, opened 2020-05-08
Add release script

fabriziocucci:
First step to automate release procedure.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -euxo pipefail
+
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
fornellas, line 1: `-e`

otherwise it keeps going on failures.

fabriziocucci: Added `-euxo pipefail` on the second line! ;)

PR #191, opened 2020-05-11
Add missing copyright

fabriziocucci:


release.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -euxo pipefail
```
fornellas, line 3: Perhaps move this after the (c) message.

PR #196, opened 2020-06-29
[mock_callable] Do not load the context on inspect

david-caro:
This speeds up considerably (from ~0.2ms to <0.01) the mocking when the
modules under test are many or the disk is slow as it skips having to
read the file to get the surrounding lines.

### DIFF
--- a/tests/lib_testslide.py
+++ b/tests/lib_testslide.py
@@ -12,6 +12,9 @@ from testslide import StrictMock
 import unittest.mock
 
 
+T = TypeVar("T")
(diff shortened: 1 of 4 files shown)


david-caro:
Locally I'm getting that error even with latest master :S

david-caro:
I think I found the issue, and sorted it out, though it's kinda tricky, so let me know what you think, thanks!!

david-caro:
Dammit, should read the comment completely first xd

PR #204, opened 2020-07-03
[global] don't load context when using getframeinfo

david-caro:
We don't really need the context (surrounding lines of code), and that
blocks the async loop and slows down tests as it has to read the
contents of the file from disk.


fornellas:
<3

PR #211, opened 2020-07-13
[lib] Fix forward references type checking

### LINKED ISSUE #208: Type validation fails with Optional quoted class
This should work:

```python
import testslide

import typing


class Foo:
    @classmethod
    def get_foo(cls) -> typing.Optional["Foo"]:
        return cls()


class Repro(testslide.TestCase):
    def test_repro(self):
        foo_mock = testslide.StrictMock(template=Foo)
        self.mock_callable(Foo, "get_foo").to_return_value(foo_mock)
        self.assertEqual(
            foo_mock,
            Foo.get_foo()
        )
```

but fails with:

```
foo_test.Repro
  test_repro: NameError: name 'Foo' is not defined

Failures:

  1) foo_test.Repro: test_repro
    1) NameError: name 'Foo' is not defined
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/contextlib.py", line 120, in __exit__
        next(self.gen)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
        yield
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 676, in run
        self._callTestMethod(testMethod)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
        method()
      File "foo_test.py", line 18, in test_repro
        Foo.get_foo()
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/site-packages/typeguard/__init__.py", line 593, in check_type
        checker_func(argname, value, expected_type, memo)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/site-packages/typeguard/__init__.py", line 410, in check_union
        check_type(argname, value, type_, memo)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/site-packages/typeguard/__init__.py", line 588, in check_type
        expected_type = resolve_forwardref(expected_type, memo)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/site-packages/typeguard/__init__.py", line 166, in resolve_forwardref
        return evaluate_forwardref(maybe_ref, memo.globals, memo.locals)
      File "/home/fornellas/.pyenv/versions/3.8.2/lib/python3.8/typing.py", line 518, in _evaluate
        eval(self.__forward_code__, globalns, localns),
      File "<string>", line 1, in <module>
        

Finished 1 example(s) in 0.5s: .
  Failed: 1
```

IIRC, there's some PEP that tells that `typing.Optional["Foo"]` is perfectly valid, but it seems that TestSlide / typeguard is / are not correctly dealing with it.

david-caro:
Fixes #208

**What:**
Pass the caller globals and locals (frame information) to typeguard to properly resolve forward references.

**Why:**
It turns out that typeguard will only check the previous frame in the
stack to extract the variables that it will use when resolving forward
references.

**How:**
Now we are extracting the globals and locals of the caller's frame when trying to resolve types, currently a bit hackish as it's done by comparing the module's path to the current module path (testslide) and testslide module's path.

**Risks:**
It might break if it's unable to extract the caller's module path.

**Checklist**:

### DIFF
--- a/tests/lib_testslide.py
+++ b/tests/lib_testslide.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import inspect
-from typing import Type, TypeVar
+from typing import Type, TypeVar, Optional
 from testslide.dsl import context, xcontext, fcontext, Skip  # noqa: F401
 import testslide.lib
 from . import sample_module
@@ -13,6 +13,12 @@ import unittest.mock
 
 
 T = TypeVar("T")
+
+
+class Foo:
+    @classmethod
+    def get_maybe_foo(cls) -> Optional["Foo"]:
+        return cls()
 
 
 @context("_validate_callable_arg_types")
@@ -334,3 +340,18 @@ def _validate_return_type(context):
     @context.example
     def fails_for_mock_with_wrong_template(self):
         self.assert_fails(StrictMock(template=int))
+
+    @context.example
+    def passes_for_valid_forward_reference(self):
+        testslide.lib._validate_return_type(
+            Foo.get_maybe_foo, Foo(), self.caller_frame_info
+        )
+
+    @context.example
+    def fails_for_valid_forward_reference_but_bad_type_passed(self):
+        with self.assertRaisesRegex(
+            TypeError, "type of return must be one of .*; got int instead:"
+        ):
+            testslide.lib._validate_return_type(
+                Foo.get_maybe_foo, 33, self.caller_frame_info
+            )
--- a/testslide/lib.py
+++ b/testslide/lib.py
@@ -3,12 +3,15 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-import inspect
 import functools
-import typeguard
-from typing import Any, Callable, Dict, Optional, Type, Tuple
-
+import inspect
+import os
+import sys
 import unittest.mock
+from typing import Any, Callable, Dict, Optional, Tuple, Type
+
+import typeguard
+
 
 ##
 ## Type validation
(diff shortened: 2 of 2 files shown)


tests/lib_testslide.py
```diff
@@ -334,3 +334,14 @@ def fails_for_wrong_type(self):
     @context.example
     def fails_for_mock_with_wrong_template(self):
         self.assert_fails(StrictMock(template=int))
+
+    @context.example
+    def passes_for_valid_forward_reference(self):
+        class Foo:
+            @classmethod
+            def get_maybe_foo(cls) -> Optional["Foo"]:  # noqa: F821 undefined name Foo
```
fornellas, line 342: Do we still need this noqa if this whole class is moved to the module level?

tests/lib_testslide.py
```diff
@@ -334,3 +334,14 @@ def fails_for_wrong_type(self):
...
+                return cls()
+
+        testslide.lib._validate_return_type(
```
fornellas, line 345: This test is dependent on the extra caller frame info argument, shouldn't we update all _validate_return_type calls everywhere to pass this argument as well?

fornellas:
Is there any way you can think of achieving this without depending on this private/protected typeguard interface: `typeguard._TypeCheckMemo`?

david-caro:
Yep, we can try passing the locals and globals, looking

On 14 July 2020 12:12:20 CEST, Fabio Pugliese Ornellas <notifications@github.com> wrote:
>Is there any way you can think of achieving this without depending on
>this private/protected typeguard interface: `typeguard._TypeCheckMemo`?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/facebookincubator/TestSlide/pull/211#issuecomment-658096242

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tests/lib_testslide.py
david-caro: `_validate_return_type` has already the `caller_frame_info` as mandatory argument:
```
def _validate_return_type(template, value, caller_frame_info):
...
```
Can you elaborate?

tests/lib_testslide.py
david-caro: Nope, though personally I prefer having it in the test that uses it (specially if no other test uses it), I'll move out though, no strong opinion.

tests/lib_testslide.py
fornellas: Oh, sorry, nevermind, I confused this with another function, this is fine :-D

tests/lib_testslide.py
```diff
@@ -334,3 +340,9 @@ def fails_for_wrong_type(self):
     @context.example
     def fails_for_mock_with_wrong_template(self):
         self.assert_fails(StrictMock(template=int))
+
+    @context.example
+    def passes_for_valid_forward_reference(self):
+        testslide.lib._validate_return_type(
+            Foo.get_maybe_foo, Foo(), self.caller_frame_info
+        )
```
fornellas, line 348: We could also cover this case:

```python
    @context.example
    def fails_for_bad_forward_reference(self):
        with self.assertRaisesWithRegex(TypeError, "int is invalid"): 
            testslide.lib._validate_return_type(
                Foo.get_maybe_foo, 33, self.caller_frame_info
            )
```

PR #215, opened 2020-07-17
Surface type errors

### LINKED ISSUE #212: Surface typing errors
We currently raise plain `TypeError` when a mock uses bad typing. If the code under test swallows this exception, the test goes green, when it should go red.

This can be accomplished by instead of raising plain `TypeError`, raise another `RuntimeTypeError` (or whatever name) that inherits from `BaseException`. This is the mechanism we already use for all mock exceptions, and it works well, as it there's virtually no expectations that code under test catches `BaseException`.

fornellas:
When runtime typing errors are detected raise `RuntimeTypeError(BaseException)` instead of plain `TypeError` to prevent code being tested to catch the exception, letting it surface to the test runner.

Closes #212 .

### DIFF
--- a/docs/patching/mock_callable/index.rst
+++ b/docs/patching/mock_callable/index.rst
@@ -407,7 +407,7 @@ This is particularly helpful when changes are introduced to the code: if a mocke
 Type Validation
 ---------------
 
-If typing annotation information is available, ``mock_callable()`` validates types of objects passing through the mock. If an invalid type is detected, it will raise ``TypeError``.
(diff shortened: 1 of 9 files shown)


david-caro:
nit: Would you consider something like `TypeCheckError` instead? It feels more descriptive to me

fornellas:
I'm changing the exception name, agree.

isort is neat, I've always missed this in black :-D

fornellas:
@david-caro  I just updated the docs, maybe you want to have a quick look before I merge it.

PR #223, opened 2020-08-07
[testslide-snippets] Add vscode snippets extension

david-caro:
DO NOT MERGE: still under development

**What:**

This adds the code for a VSCode extension adding some helpful snippets
to develop with TestSlide.

**Why:**

This will help develop tests faster, improving considerably the developer experience (in most cases).

**Checklist**:


cricalix:
Some thoughts:

1. All of the examples are for non-async methods. I was going to suggest tsa, but tsa is already taken as a prefix.
2. It may make sense to use the TM_SELECTED_TEXT fragment in things like the snippet for creating an example. Someone might be porting an existing block of code, and if they've selected some text - Ctrl-Shift-P, Insert Snippet<enter>, arrow key to relevant snippet - will let them just wrap the selection. It takes a default, so you can do ${TM_SELECTED_TEXT:default_if_no_selection}.
3. From the documented nomenclature, tsb_full should be tsab_full, because it has before and after?

util/testslide-snippets/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
+## 1.0.0
```
fornellas, line 2: @david-caro  do you think it makes sense to sync this with TestSlide's version? If yes, we should also update `RELEASE.md` to cover this.

util/testslide-snippets/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+The snippets follow some common guidelines:
+
+- All snippets are prefixed with `ts` (for TestSlide).
+- Snippets starting with `tsb_` will generate basic test cases and add a before clause.
+- Snippets starting with `tsa_` will generate basic test cases and add an after clause.
+- Snippets starting with `tsab_` will generate basic test cases with both before and after clause.
+- Snippets starting with `tsab_` will generate basic async test cases and add a before clause.
+- Snippets starting with `tsaa_` will generate basic async test cases and add an after clause.
+- Snippets starting with `tsaab_` will generate basic async test cases with both before and after clause.
```
fornellas, line 15: I think this is bound to get out of sync with snippets. How about just keep the "everything is ts*"?

util/testslide-snippets/CHANGELOG.md
david-caro: Done

fornellas:
This is good to merge, right @david-caro ?

david-caro:
Now it is, added the note on how to publish the extension :)

david-caro:
The coverage seems the same, the report is unable to tell me what changed exactly

Issue #159, opened 2020-04-14
Add type validation to `patch_attribute()`


fornellas:
Similarly to what we already have for StrictMock, mock_callable, mock_async_callable & mock_constructor.

Test cases that needs covering:

- Modules.
- Classes.
- Instances of classes.
- Class attributes at instances of classes.
- Properties at instances of classes.

deathowl:
closed by #227

PR #227, opened 2020-08-23
implement type validation for patch_attribute

deathowl:
**What:**

Add type validation to patch_attribute

**Why:**
#159 

**Checklist**:


docs/patching/patch_attribute/index.rst
```diff
@@ -41,3 +41,8 @@ Properties are tricky to patch because of the quirky mechanics that `Python's De
       with_property = WithProperty()
       self.patch_attribute(with_property, "prop", "mock")
       self.assertEqual(with_property.prop, "mock")
+
+Type Validation
+---------------
+
+patch_attribute implements type validation. When you use it, the mock will raise ``TypeCheckerError`` if it is called with a type that does not match the original attribute type. This feature can be disabled with ``type_validation=False``
```
fornellas, line 48: I think this phrase would be fit for mock_callable: patch_attribute works exclusively for non-callable attributes.

Suggesstion: `patch_attribute` implements type validation and will only allow attributes to be mocked with values of types that match the available typing annotation. `TypeCheckerError` will be raised if a bad type is given. This check is expected to work 100% of times, but if it does not (possibly a  bug in TestSlide, please report!), you can disable it by doing `patch_attribute(obj, "attr", type_validation=False)`.

tests/patch_attribute_testslide.py
```diff
@@ -110,9 +111,17 @@ def it_fails_if_attribute_is_a_class(self):
             def with_class_attributes(context):
                 context.merge_context("patching works")
 
-        @context.xexample
+        @context.example
         def it_fails_if_new_value_is_of_incompatible_type(self):
-            pass
+            with self.assertRaises(TypeCheckError):
+                print(self.target)
```
fornellas, line 117: Drop

tests/patch_attribute_testslide.py
```diff
@@ -110,9 +111,17 @@ def it_fails_if_attribute_is_a_class(self):
...
+                self.patch_attribute(self.target, "typedattr", 123)
+
+        @context.example
+        def it_passes_if_new_value_is_of_incompatible_type_with_type_validation_false(
+            self,
+        ):
+            self.patch_attribute(self.target, "typedattr", 123, type_validation=False)
```
fornellas, line 124: I think we should also add a positive test case "It patches & works with valid type".

testslide/patch_attribute.py
```diff
@@ -56,6 +60,7 @@ def patch_attribute(
     key = (id(target), attribute)
 
     if isinstance(target, testslide.StrictMock):
+        target.__dict__["_type_validation"] = type_validation
```
fornellas, line 63: I believe this will change this StrictMock instance constructor parameter on the fly:

- StrictMock is created with "type_validation=True".
- We do `patch_attribute(strict_mock, "att", type_validation=False)`
- StrictMock instance is now has no type validation.

What I would expect, is for `patch_attribute(strict_mock, "att", type_validation=False)` to only affect the attribute being patched. I suppose you added this here, because otherwise, despite `patch_attribute(strict_mock, "att", type_validation=False)`, StrictMock would still type check this attribute later.

I suppose a way out, would be for StirctMock to have some private attribute like `_attributes_to_skip_type_validation`, and have `patch_attribute` update it instead.

Makes sense?

fornellas: PS: Either way, adding a test case specifically to cover patch_attribute/StrictMock integration is a good idea :-D

deathowl: Good idea, will do that.

docs/patching/patch_attribute/index.rst
```diff
@@ -41,3 +41,8 @@ Properties are tricky to patch because of the quirky mechanics that `Python's De
       with_property = WithProperty()
       self.patch_attribute(with_property, "prop", "mock")
       self.assertEqual(with_property.prop, "mock")
+
+Type Validation
+---------------
+
+``patch_attribute`` implements type validation and will only allow attributes to be mocked with values of types that match the available typing annotation. TypeCheckerError will be raised if a bad type is given. This check is expected to work 100% of times, but if it does not (possibly a bug in TestSlide, please report!), you can disable it by ``doing patch_attribute(obj, "attr", "newval", type_validation=False)``.
```
fornellas, line 48:
```
``TypeCheckerError``
```

tests/patch_attribute_testslide.py
```diff
@@ -110,9 +111,20 @@ def it_fails_if_attribute_is_a_class(self):
...
+
+        @context.example
+        def it_passes_if_new_value_is_of_matching_type(self,):
+            self.patch_attribute(self.target, "typedattr", "mocked")
```
fornellas, line 127: Suggestion: group all type validation tests under the same context:

```
@context.sub_context
def type_validation(context):
  @context.example
  ...
```

deathowl: done.

docs/patching/patch_attribute/index.rst
deathowl: done.

testslide/strict_mock.py
```diff
@@ -459,6 +460,7 @@ def __init__(
         name=None,
         default_context_manager=False,
         type_validation=True,
+        attributes_to_skip_type_validation=[],
```
fornellas, line 463: Documentation missing:

- At this method's doc string
- docs/strict_mock/...

fornellas (approved):
LGTM, just please add documentation for StrictMock's new constructor parameter.

docs/strict_mock/index.rst
```diff
@@ -542,8 +542,32 @@ If this type validation yields bad results (likely a bug, please report it), you
 
   StrictMock(template=SomeClass, type_validation=False)
 
+Attribute Type Validation
+=========================
```
fornellas, line 546: This can be a sub section of "Type Validation"

```
Per Attribute Type Validation
-----------------------------
```

docs/strict_mock/index.rst
```diff
@@ -542,8 +542,32 @@ If this type validation yields bad results (likely a bug, please report it), you
...
+By default, ``StrictMock`` will validate types of set attributes, method call arguments and method return values, against available type hinting information.
+If you want to disable type validation for certain attributes pass ``attributes_to_skip_type_validation`` to the constructor of ``StrictMock``
```
fornellas, line 548: `type_validation=False` works in combination with `attributes_to_skip_type_validation`: either global or per attribute. Perhaps it is worth explaining this, maybe merge this section with the previous section (they're kind of the same).

deathowl: Sorry I am bad at writing docs :D

PR #230, opened 2020-08-26
Fix template validation

fornellas:
We were not raising `ValueError` for bad templates when `bool(template) == False`.


deathowl:
LGTM, just make format :)

deathowl (approved):

fornellas:
Dammit, I had an outdated black version. I'm considering adding a `Makefule` rule like:

```
python -m venv --clear venv && bash -c 'source venv/bin/activate && make install_build_deps && make travis'
```
so we can get signal closer to Travis', what do you think @deathowl ?

Issue #156, opened 2020-04-09
Disallow re-setting attributes


fornellas:
Currently from the DSL users can arbitrarily re-set context attributes:

```
@context
def top(context):
  @context.before
  def first_before(self):
    self.base = "base"
    self.derived = f"derived: {self.base}"

  @context.sub_context
  def inner(context):
    @context.before
    def second_before(self):
      self.base = "new base"
    @context.example
    def composition(self):
      self.assertEqual(self.derived, "derived: new base")  # Not True!
```

This is bad, as it prevents composition to work (eg: a sub context define a new value for an attribute).

We can change this, so that context attributes can be set **only once**.

This can probably be somewhat easily implemented by defining `_ContextData.__setattr__` with the "set once" rule.

deathowl:
@fornellas 
I added a naive implementation of this on my local fork
```
diff --git a/testslide/__init__.py b/testslide/__init__.py
index e013bd7..8254e81 100644
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -115,6 +115,15 @@ class _ContextData(object):
     def _all_memoizable_attributes(self):
         return self._context.all_context_data_memoizable_attributes

+    def __setattr__(self, name, value):
+        if self.__dict__.get(name) and self.__dict__[name] != value:
+            raise AttributeError(
+                f"Attribute: {name} is already set.\
+        Resetting attributes in sub-contexts is not allowed, as it breaks composition"
+            )
+        else:
+            super(_ContextData, self).__setattr__(name, value)
+
     def __getattr__(self, name):
         if name in self._all_methods.keys():

```

However, it breaks current testslide test suite, since we're resetting a couple of attributes in sub-contexts :
```
➜  TestSlide git:(disallow_resetting_attributes) ✗ make
COVERAGE ERASE
UNITTEST tests/cli_unittest.py
UNITTEST tests/dsl_unittest.py
UNITTEST tests/matchers_unittest.py
UNITTEST tests/testcase_unittest.py
TESTSLIDE tests/lib_testslide.py
TESTSLIDE tests/mock_async_callable_testslide.py
mock_async_callable()
  when target is a module
    and callable is a function
      can not mock
      can mock with flag
    and callable is an async function
      default behavior
      .for_call()
      .to_raise(exception)
      .to_call_original()
      .and_assert_*
      .to_return_value(value)
        it returns value
        return value type
          passes with valid type
          with invalid return type
            raises TypeCheckError
[...]

PR #237, opened 2020-09-07
Disallow resetting attributes

deathowl:
**What:**
Disallow resetting attributes

**Why:**

#156 

**How:**
Make attributes set once

**Risks:**

**Checklist**:

### DIFF
--- a/tests/mock_callable_testslide.py
+++ b/tests/mock_callable_testslide.py
@@ -605,15 +605,15 @@ def mock_callable_tests(context):
 
             @context.shared_context
             def integration(context):
-                @context.before
-                def catch_callable_target_exceptions(self):
-                    original_callable_target = self.callable_target
+                @context.memoize_before
+                def callable_target(self):
+                    original_callable_target = self._original_target
 
                     def _callable_target(*args, **kwargs):
                         with self.assertRaises(self.exception_class):
                             return original_callable_target(*args, **kwargs)
 
-                    self.callable_target = _callable_target
+                    return _callable_target
 
                 if has_original_callable:
                     context.merge_context("mock call arguments", has_return_value=False)
@@ -1111,6 +1111,10 @@ def mock_callable_tests(context):
             def callable_target(self):
                 return getattr(self.real_target, self.callable_arg)
 
+            @context.memoize_before
+            def _original_target(self):
+                return getattr(self.real_target, self.callable_arg)
+
             context.merge_context("mock configuration examples")
 
         @context.sub_context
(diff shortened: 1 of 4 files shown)


deathowl:
Some Strcitmock related cases still fail in the mock_callable suite. As discussed with @fornellas  he will help in those.

fornellas:
LGTM. I'm gonna be a bit busy in the following days, but after, I can give a hand with updating the tests.

testslide/__init__.py
```diff
@@ -115,6 +115,16 @@ def _all_methods(self):
     def _all_memoizable_attributes(self):
         return self._context.all_context_data_memoizable_attributes
 
+    def __setattr__(self, name, value):
+        if self.__dict__.get(name) and self.__dict__[name] != value:
+            raise AttributeError(
+                f"Attribute: {name} is already set.\
+        Resetting attributes in sub-contexts is not allowed, as it breaks composition.\
+            Use memoize and composition instead"
+            )
```
fornellas, line 124: This is a non-trivial thing to understand for the average users. I suggest giving a more thorough and detailed message here.

PS: I think the white spaces from code formatting were part of the string.

```
                f"Attribute {repr(name)} is already set.\n"
                "Changing the value of attributes after they have been set "
                "can lead to unexpected test results. Eg: when a sub context "
                "resets an attribute after a parent context has set and used "
                "it, they will have different objects for the same attribute.\n"
                "You can safely override attributes from parent contexts by "
                "using @context.before, @context.memoize_before or "
                "@context.function, so the inner-most definition is used."
```

tests/mock_callable_testslide.py
```diff
@@ -605,15 +605,15 @@ def to_raise_exception(context):
 
             @context.shared_context
             def integration(context):
-                @context.before
-                def catch_callable_target_exceptions(self):
-                    original_callable_target = self.callable_target
+                @context.memoize_before
```
fornellas, line 608: Oh, this was tricky... I already regret having written it as it is, lol.

tests/mock_callable_testslide.py
```diff
@@ -1111,6 +1111,10 @@ def mock_callable_dsl(self):
             def callable_target(self):
                 return getattr(self.real_target, self.callable_arg)
 
+            @context.memoize_before
+            def _original_target(self):
+                return getattr(self.real_target, self.callable_arg)
+
```
fornellas, line 1117: Nit: I'd move this to line 607-608 so it is defined before it is referred for easier reading.

fornellas (approved):
LGTM.

I was thinking we could add a doc section under this one
https://testslide.readthedocs.io/en/master/testslide_dsl/context_attributes_and_functions/index.html#attributes
giving an example of resetting attributes, and how parent / child context may see different values and this being an issue, hooking to the next section of memoized attributes as the solution.

PR #241, opened 2020-09-12
Fix mock callable/constructor to not accept extra configuration after…

ldfsilva:
**What:**

This fixes [issue 16](https://github.com/facebookincubator/TestSlide/issues/16), not allowing that extra configuration is made after `mock_callable`, `mock_async_callable` or mock_constructor` receive their first call.

Added tests to cover the constraint to not allow extra configuration to be made in `mock_callable`, `mock_async_callable` and `mock_constructor` and also validated with ad-hoc tests that *TestSlide* tests continues to behave as expected. As show below:

For the ad-hoc tests I used this [test_extra_configuration.py](https://gist.github.com/ldfsilva/b4b6b7964ef9cda7819059569997efbb) and invoked tests as follows:

```
python -m unittest -v tests.test_extra_configuration
```

Here are excerpts from the test results: [[here](https://gist.github.com/ldfsilva/d88ff7e16964080d16e59b2510da21fc)]

**mock_async_calable:**
```
ERROR: test_mock_async_callable_fail_already_received_1_call (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_async_callable receives its first call, it received 1 call already. You should instead define it all at once, eg: self.mock_async_callable(target, 'func').to_return_value(value).and_assert_called_once()
```

```
ERROR: test_mock_async_callable_fail_already_received_8_calls (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_async_callable receives its first call, it received 8 calls already. You should instead define it all at once, eg: self.mock_async_callable(target, 'func').to_return_value(value).and_assert_called_once()
```

**mock_calable:**

```
ERROR: test_mock_callable_fail_already_received_1_call (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_callable receives its first call, it received 1 call already. You should instead define it all at once, eg: self.mock_callable(target, 'func').to_return_value(value).and_assert_called_once()
```

```
ERROR: test_mock_callable_fail_already_received_8_calls (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_callable receives its first call, it received 8 calls already. You should instead define it all at once, eg: self.mock_callable(target, 'func').to_return_value(value).and_assert_called_once()
```

**mock_constructor:**
```
ERROR: test_mock_constructor_fail_already_received_1_call (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_constructor receives its first call, it received 1 call already. You should instead define it all at once, eg: self.mock_constructor(target, 'func').to_return_value(value).and_assert_called_once()
```

```
ERROR: test_mock_constructor_fail_already_received_8_calls (tests.test_extra_configuration.TestCase)
ValueError: No extra configuration is allowed after mock_constructor receives its first call, it received 8 calls already. You should instead define it all at once, eg: self.mock_constructor(target, 'func').to_return_value(value).and_assert_called_once()
```

### DIFF
--- a/testslide/mock_callable.py
+++ b/testslide/mock_callable.py
@@ -572,10 +572,11 @@ class _CallableMock(object):
 class _MockCallableDSL(object):
 
     CALLABLE_MOCKS: Dict[int, _CallableMock] = {}
+    _NAME: str = "mock_callable"
 
     def _validate_patch(
         self,
-        name="mock_callable",
+        name=_NAME,
         other_name="mock_async_callable",
         coroutine_function=False,
         callable_returns_coroutine=False,
@@ -770,6 +771,15 @@ class _MockCallableDSL(object):
                 "self.mock_callable(target, 'func')"
                 ".to_return_value(value)"
                 ".and_assert_called_exactly(times)"
+            )
+        if self._runner._call_count > 0:
+            raise ValueError(
+                f"No extra configuration is allowed after {self._NAME} "
+                f"receives its first call, it received {self._runner._call_count} "
+                f"call{'s' if self._runner._call_count > 1 else ''} already. "
+                "You should instead define it all at once, "
+                f"eg: self.{self._NAME}(target, 'func')"
+                ".to_return_value(value).and_assert_called_once()"
             )
 
     ##
@@ -1005,6 +1015,9 @@ class _MockCallableDSL(object):
 
 
 class _MockAsyncCallableDSL(_MockCallableDSL):
+
+    _NAME: str = "mock_async_callable"
+
     def __init__(
         self,
         target,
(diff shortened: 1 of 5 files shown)


deathowl:
LGTM. @ldfsilva  Just fix the breaking testcase plase :)

ldfsilva:
sounds good, I just noticed that we have a designated test file for `mock_async_callable` (need glasses), so I relocated that test over there, removed `asyncio.run` and fixed linter, it should be good now.

testslide/mock_callable.py
```diff
@@ -572,10 +572,11 @@ def _registered_calls(self):
 class _MockCallableDSL(object):
 
     CALLABLE_MOCKS: Dict[int, _CallableMock] = {}
+    _NAME: str = "mock_callable"
 
     def _validate_patch(
         self,
-        name="mock_callable",
+        name=_NAME,
```
fornellas, line 579: I think we can drop the name argument here, as it can (should!) be accessed with `self._NAME`.

fornellas:
Thanks @ldfsilva  for working on this! Small, focused and with awesome simple test coverage!

PR #220, opened 2020-07-31
Add TestSlide logo to README and docs

lhl2617:
**What:**

Added logo for TestSlide in `README.md` and docs

**Why:**

TestSlide now has a logo!

**How:**

Added `docs/testslide_logo.png`, and references to it from `README.md` and `docs/index.rst`

**Risks:**

N/A

**Checklist**:

- [x] Added tests, if you've added code that should be tested
- [x] Updated the documentation, if you've changed APIs
- [x] Ensured the test suite passes
- [x] Made sure your code lints
- [x] Completed the Contributor License Agreement ("CLA")

PR #243, opened 2020-09-28
Add logo assets from #220

fornellas:
Let's add the logo assets to the repo so it is easier to find than an attached zip at a GH issue.


ldfsilva (approved):
awesome, that'll will help with discovery, also for linking externally 👍

PR #245, opened 2020-10-01
Detect leaked tasks

### LINKED ISSUE #244: Catch started but not finished tasks
**I'm using `TestSlide` version:** 2.5.9

**Given:** Code that creates an asyncio task but does not wait for it (see below):

```python
#!/usr/bin/env python3
import asyncio
import sys


def to_mock():
    pass


async def some_async_function():
    to_mock()


@context
def async_test(context):
    @context.example
    async def test_assert_called_fails(self):
        self.mock_callable(sys.modules[__name__], 'to_mock')\
            .to_return_value(None)\
            .and_assert_called_once()
        asyncio.create_task(some_async_function())

    @context.example
    async def test_assert_called_works_when_extra_await(self):
        self.mock_callable(sys.modules[__name__], 'to_mock')\
            .to_return_value(None)\
            .and_assert_called_once()
        asyncio.create_task(some_async_function())
        await asyncio.sleep(1)
```

**When I run:** testslide example.py

```
 # testslide example.py 
async test
  test assert called fails: AssertionError: calls did not match assertion.
  test assert called works when extra await

Failures:

  1) async test: test assert called fails
    1) AssertionError: calls did not match assertion.
    <module 'example' from '/home/dcaro/example.py'>, 'to_mock':
      expected: called exactly 1 time(s) with any arguments   received: 0 call(s)
      File "work/TestSlide/testslide/mock_callable.py", line 250, in assertion
        raise AssertionError(

Finished 2 example(s) in 2.1s 
  Successful: 1
  Failed: 1
```

**I expected this to happen:** Testslide should tell me that there were non-finished tasks when the test finished, maybe just a warning.

**But, instead this happened:** Testslide just fails as if the function was not called (as it wasn't, because the task ran/was cancelled after the test finished).

david-caro:
**What:**

Add leaked tasks detection to both sync and async tests.
Fixes #244

**Why:**

If you were to start a task (ex. using asyncio.create_task) and not await for it, there would be no indication that the task was finished (or not) before the test ended.

**How:**

Wrap the execution of the test (the async block of code specifically) and check the tasks that were running before and after, and if there's any new task running after then throw an exception.

**Risks:**

Tests were not checking this issue before, so any test that before might have been leaking tasks, will not start failing.

**Checklist**:

### DIFF
--- a/docs/testslide_dsl/async_support/index.rst
+++ b/docs/testslide_dsl/async_support/index.rst
@@ -43,7 +43,7 @@ like this:
       assert self.memoize_before == "memoize_before"
       assert self.function == "function"
 
-The test runner will create a new event look to execute each example.
+The test runner will create a new event loop to execute each example.
 
 .. note::
 
@@ -184,3 +184,36 @@ Python's default threshold for triggering this event loop lock up failure is **1
 
   Finished 1 example(s) in 1.0s
     Successful: 1
+
+Leaked Tasks
+^^^^^^^^^^^^
+
+If your async code creates a task in the asyncio loop, but finished before that task has ended (ex. you forgot to await for it), testslide will catch it and fail the test.
+
+This is enabled by default for async tests, but to get that behaviour also when running async code from sync tests, for example:
+
+
+.. code-block:: python
+
+  import asyncio
+  from testslide.dsl import context
+
+  @context
+  def my_test_suite(context):
+        @context.example
+        def test_something_async(self):
+            asyncio.run(my_async_function())
+
+
+Has to use the `async_run` function from the context, so instead, you should use:
+
+.. code-block:: python
+
+  import asyncio
+  from testslide.dsl import context
+
+  @context
+  def my_test_suite(context):
+        @context.example
+        def test_something_async(self):
+            self.async_run(my_async_function())
--- a/docs/testslide_dsl/async_support/index.rst
+++ b/docs/testslide_dsl/async_support/index.rst
@@ -205,7 +205,7 @@ This is enabled by default for async tests, but to get that behaviour also when 
             asyncio.run(my_async_function())
 
 
-Has to use the `async_run` function from the context, so instead, you should use:
+Has to use the `async_run_with_health_checks` function from the context, so instead, you should use:
 
 .. code-block:: python
 
@@ -216,4 +216,4 @@ Has to use the `async_run` function from the context, so instead, you should use
   def my_test_suite(context):
         @context.example
         def test_something_async(self):
-            self.async_run(my_async_function())
+            self.async_run_with_health_checks(my_async_function())
--- a/tests/dsl_unittest.py
+++ b/tests/dsl_unittest.py
@@ -2073,13 +2073,63 @@ class TestMockConstructorIntegration(TestDSLBase):
                     ).to_return_value("mocked_popen").and_assert_called_once()
                     assert subprocess.Popen(["cmd"]) == "mocked_popen"
 
-        examples = {}
-
-        for all_top_level_context in Context.all_top_level_contexts:
-            for example in all_top_level_context.all_examples:
-                examples[example.name] = example
+        examples = _get_name_to_examples()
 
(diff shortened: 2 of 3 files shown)


deathowl (approved):

docs/testslide_dsl/async_support/index.rst
```diff
@@ -184,3 +184,36 @@ Python's default threshold for triggering this event loop lock up failure is **1
 
   Finished 1 example(s) in 1.0s
     Successful: 1
+
+Leaked Tasks
+^^^^^^^^^^^^
+
+If your async code creates a task in the asyncio loop, but finished before that task has ended (ex. you forgot to await for it), testslide will catch it and fail the test.
+
+This is enabled by default for async tests, but to get that behaviour also when running async code from sync tests, for example:
```
fornellas, line 193: Shouldn't `self.async_run`  encapsulate all event loop health checks we have, and not only the leaked tasks?

tests/dsl_unittest.py
```diff
@@ -24,6 +24,15 @@
 from testslide.runner import QuietFormatter
 
 
+def _get_name_to_examples():
```
fornellas, line 27: <3

tests/dsl_unittest.py
```diff
@@ -2023,17 +2024,31 @@ async def expect_pass(self):
...
 
+    def test_leaked_tasks_fails_test_run(self):
+        async def dummy_async_func():
+            pass
+
+        @context
+        def fail_top(context):
+            @context.example
+            async def spawn_task_but_dont_await(self):
+                # once we stop supporting python 3.6 we can use
```
fornellas, line 2042: Can we have an `if python version <= 3.6` here please? It helps cleanup in the future, as we can merely delete code, instead of adding.

testslide/__init__.py
```diff
@@ -50,6 +74,21 @@ def dot_lookup(thing, comp, import_path):
...
 
 
+async def _async_ensure_no_leaked_tasks(coro):
+    before_example_tasks = get_all_tasks()
+    result = await coro
+    after_example_tasks = get_all_tasks()
+    new_still_running_tasks = set(after_example_tasks) - set(before_example_tasks)
+    if new_still_running_tasks:
+        tasks_str = "\n".join(str(task) for task in new_still_running_tasks)
+        raise RuntimeError(
```
fornellas, line 84: Other asyncio health issues raise their own `BaseException` classes. Reason for this, is to avoid the scenario where this exception is raised under code being tested such as:

```
try:
  do_clowny_async_thing()
except Exception:
  ...
```

and the tested code swallows the exception, tests go green, but are actually broken.

Can we have something like `class LeftOverActiveTasks(BaseException):` instead?

tests/dsl_unittest.py
```diff
@@ -2058,13 +2073,63 @@ def expect_pass(self):
...
-        examples = {}
+        examples = _get_name_to_examples()
+
+        self.run_example(examples["expect pass"])
+
+        with self.assertRaisesRegex(AssertionError, "calls did not match assertion"):
+            self.run_example(examples["expect fail"])
+
+
+class TestMock(TestDSLBase):
```
fornellas, line 2084: Isn't this a duplicate of `TestMockConstructorIntegration` above?

testslide/__init__.py
```diff
@@ -175,6 +214,15 @@ def sub_example(self, name=None):
         with self._sub_examples_agg_ex.catch():
             yield
 
+    def async_run(self, coro):
```
fornellas, line 217: If we think about TestSlide test runner being able to also run `unittest.TestCase` tests, it is far to argue that having asyncio health checks there would be cool as well, so maybe we should also have it at `testslide.TestCase`.

Also, I can picture users asking "why use self.asyncio_run and not asyncio.run?", so I'd consider calling this method something like `asyncio_run_with_health_checks`  (or whatever).

fornellas:
LGTM, just some small nits :-D

PR #246, opened 2020-10-02
Mega typechecking PR

deathowl:
**What:**

Add strict mypy rules to testslide code. Make those strict mypy checks pass, while retaining functional state of testslide

**Why:**

Typing is a nice thing

**How:**

partially using https://github.com/Instagram/MonkeyType, fixing the resulting errors (lacking forward references, and inferred types from test code) and add typing info to functions that missed the sampling during tests

**Risks:**

Tests still pass :)

**Checklist**:

PLS REVIEW THOROUGHLY

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,3 +4,9 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-psutil.*]
 ignore_missing_imports = True
+[mypy-testslide.testslide]
+disallow_untyped_calls = True
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
+disallow_untyped_decorators = True
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -21,6 +21,7 @@ import unittest
 import warnings
 from contextlib import contextmanager
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     Dict,
@@ -39,7 +40,7 @@ import testslide.mock_constructor
 import testslide.patch_attribute
 from testslide.strict_mock import StrictMock  # noqa
 
-if False:
+if TYPE_CHECKING:
     # hack for Mypy
     from testslide.runner import BaseFormatter
 
@@ -100,7 +101,7 @@ class _ContextData(object):
         def register_assertion(assertion: Callable) -> None:
             if self._example.is_async:
 
-                async def f(_: Any) -> None:
+                async def f(_: _ContextData) -> None:
                     assertion()
 
             else:
@@ -119,7 +120,7 @@ class _ContextData(object):
         self._init_mocks()
 
     @staticmethod
-    def _not_callable(self: Any) -> None:
+    def _not_callable(self: "_ContextData") -> None:
         raise BaseException("This function should not be called outside test code.")
 
     @property
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -20,7 +20,7 @@ import types
 import unittest
 import warnings
 from contextlib import contextmanager
-from typing import List
+from typing import Any, Callable, Dict, Iterator, Optional, Type, Union, List, TextIO, Tuple
 
 import testslide.matchers
 import testslide.mock_callable
@@ -28,16 +28,21 @@ import testslide.mock_constructor
 import testslide.patch_attribute
 from testslide.strict_mock import StrictMock  # noqa
 
+if False:
+    # hack for Mypy
+    from testslide.runner import BaseFormatter
+    from unittest.mock import Mock
+
 if sys.version_info < (3, 6):
     raise RuntimeError("Python >=3.6 required.")
 
 
-def _importer(target):
+def _importer(target: str) -> Any:
     components = target.split(".")
     import_path = components.pop(0)
     thing = __import__(import_path)
 
-    def dot_lookup(thing, comp, import_path):
+    def dot_lookup(thing: object, comp: str, import_path: str) -> Any:
         try:
             return getattr(thing, comp)
         except AttributeError:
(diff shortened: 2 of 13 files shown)


mypy.ini
```diff
@@ -3,4 +3,10 @@ ignore_missing_imports = True
...
 [mypy-psutil.*]
-ignore_missing_imports = True
\ No newline at end of file
+ignore_missing_imports = True
+[mypy-testslide.testslide]
+disallow_untyped_calls = True
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
+disallow_untyped_decorators = True
```
fornellas, line 12: Bold, and GOLD!

testslide/__init__.py
```diff
@@ -20,24 +20,39 @@
...
+)
 
 import testslide.matchers
 import testslide.mock_callable
 import testslide.mock_constructor
 import testslide.patch_attribute
 from testslide.strict_mock import StrictMock  # noqa
 
+if False:
+    # hack for Mypy
```
fornellas, line 43: Why? If this is some chicken / egg issue with imports, quoting the types usually helps.

testslide/__init__.py
```diff
@@ -56,36 +71,36 @@ class _ContextData(object):
...
         if self._example.is_async:
 
-            async def assert_sub_examples(self):
+            async def assert_sub_examples(self: "_ContextData") -> None:
                 real_assert_sub_examples(self)
 
         else:
 
-            def assert_sub_examples(self):
+            def assert_sub_examples(self: "_ContextData") -> None:  # type: ignore
```
fornellas, line 88: add typing info AND ignore it?

testslide/__init__.py
```diff
@@ -56,36 +71,36 @@ class _ContextData(object):
...
         self.patch_attribute = testslide.patch_attribute.patch_attribute
-        self._mock_callable_after_functions = []
+        self._mock_callable_after_functions: List[Callable] = []
 
-        def register_assertion(assertion):
+        def register_assertion(assertion: Callable) -> None:
             if self._example.is_async:
 
-                async def f(_):
+                async def f(_: Any) -> None:
```
fornellas, line 103: I'm almost positive this `Any` is `ContextData`

testslide/__init__.py
```diff
@@ -94,28 +109,28 @@ async def f(_):
...
         self._context = example.context
-        self._after_functions = []
+        self._after_functions: List[Callable] = []
         self._test_case = unittest.TestCase()
         self._init_sub_example()
         self._init_mocks()
 
     @staticmethod
-    def _not_callable(self):
+    def _not_callable(self: Any) -> None:
```
fornellas, line 122: IIRC, Any is ContextData.

testslide/__init__.py
```diff
@@ -157,7 +172,7 @@ def static(*args, **kwargs):
                 "Context '{}' has no attribute '{}'".format(self._context, name)
             )
 
-    def after(self, after_code):
+    def after(self, after_code: Callable) -> Callable:
```
fornellas, line 175: IIRC, both should be `Callable[[ContextData], None]`.

testslide/__init__.py
```diff
@@ -181,29 +196,31 @@ class AggregatedExceptions(Exception):
     Aggregate example execution exceptions.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         super(AggregatedExceptions, self).__init__()
-        self.exceptions = []
+        self.exceptions: List[Union[Exception, BaseException]] = []
```
fornellas, line 201: `List[BaseException]` shouldn't this work? Exceptions are BaseExceptions after all...

testslide/__init__.py
```diff
@@ -181,29 +196,31 @@ class AggregatedExceptions(Exception):
...
 
-    def append_exception(self, exception):
+    def append_exception(
+        self, exception: Union[Exception, "AggregatedExceptions", BaseException]
```
fornellas, line 204: Same here, shouldn't it be BaseException only?

testslide/__init__.py
```diff
@@ -240,20 +257,24 @@ class SlowCallback(Exception):
...
-    def __init__(self, example, formatter):
+    def __init__(self, example: "Example", formatter: "BaseFormatter") -> None:
         self.example = example
         self.formatter = formatter
         self.trim_path_prefix = self.formatter.trim_path_prefix
 
     @staticmethod
-    async def _fail_if_not_coroutine_function(func, *args, **kwargs):
+    async def _fail_if_not_coroutine_function(
+        func: Callable, *args: Any, **kwargs: Any
```
fornellas, line 267: IIRC: `Callable[[ContextData, ...], Any]`.

testslide/__init__.py
```diff
@@ -240,20 +257,24 @@ class SlowCallback(Exception):
...
         if not inspect.iscoroutinefunction(func):
             raise ValueError(f"Function must be a coroutine function: {repr(func)}")
         return await func(*args, **kwargs)
 
     async def _real_async_run_all_hooks_and_example(
-        self, context_data, around_functions=None
-    ):
+        self,
+        context_data: _ContextData,
+        around_functions: Optional[List[Callable]] = None,
```
fornellas, line 276: IIRC `Callable[[ContextData, ...], Any]`

testslide/__init__.py
```diff
@@ -284,7 +305,7 @@ async def _real_async_run_all_hooks_and_example(
                 await self._fail_if_not_coroutine_function(
                     self.example.code, context_data
                 )
-            after_functions = []
+            after_functions: List[Callable] = []
```
fornellas, line 308: IIRC `Callable[[ContextData, ...], Any]`

testslide/__init__.py
```diff
@@ -466,7 +500,14 @@ class Example(object):
     Individual example.
     """
 
-    def __init__(self, name, code, context, skip=False, focus=False):
+    def __init__(
+        self,
+        name: str,
+        code: Callable,
```
fornellas, line 506: IIRC `Callable[[ContextData, ...], Any]`

testslide/__init__.py
```diff
@@ -504,47 +545,70 @@ class _TestSlideTestResult(unittest.TestResult):
...
+    ) -> None:
         exc_type, exc_value, exc_traceback = err
         self.aggregated_exceptions.append_exception(exc_value)
 
-    def addError(self, test, err):
+    def addError(  # type:ignore
+        self,
+        test: "TestCase",
+        err: Tuple[
+            Union[Type[Exception], Type[BaseException]],
```
fornellas, line 567: can't we have only BaseException here?

fornellas (approved):
@deathowl  thanks SO MUCH for cleaning up after this! This is HUGE, both in size and impact! We'll have some sanity now at this codebase :-D

I read through all changes, I spotted nothing that caught my attention other than these 2 situations, which repeated a lot:

- Unions of `BaseException` & other exception types (eg: `Exception`). Shouldn't `BaseException` be enough? All Python exception are subclasses of it, I think it should work just fine with only it.
- There are lots of `Callable` types (no args, no return) or `Callable[[a,b,c...], d]`, mostly related to DSL hooks, such as examples, before / after hooks etc. TBH, this PR is awesome as it is, but I was thinking about having custom types such as `ExampleFunction = Callable[[_ContextData], None]` (non-blocking here to be clear).

I'm letting this pre-approved, just please make sure that build passes before merging (small nits to fix only).

testslide/__init__.py
deathowl: if i just quote the type, then mypy will fail with cannot find type "asdf" if i dont "import" the depending module like so. also i can't properly import that due to circular depencendy things. :D

deathowl: will give typing.TYPE_CHECKING a try

deathowl: yep, thx

testslide/__init__.py
deathowl: lol. thx. Must have been drunk or idk :D

deathowl: it's because in one case it's sync in the other case it's async. So it causes: testslide/__init__.py:89: error: All conditional function variants must have identical signatures

PR #251, opened 2020-10-16
fix #247 make _Config class a dataclass and remote type:ignore

### LINKED ISSUE #247: Make _Config in runner.py class a dataclass so we can remove the whole lot of type: ignore-s from later parts of cli.py
This not being a dataclass causes a lots of confusion for mypy, and rightfully so. This is not the nicest. Make it a dataclass or replace everything with click :)

xiaoxianma:
**What:**

**Why:**

**How:**

**Risks:**

**Checklist**:

### DIFF
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -421,7 +421,7 @@ class _ExampleRunner:
             else:
                 original_showwarning(message, category, filename, lineno, file, line)
 
-        warnings.showwarning = showwarning
+        warnings.showwarning = showwarning  # type: ignore
 
         original_logger_warning = asyncio.log.logger.warning
 
@@ -438,7 +438,7 @@ class _ExampleRunner:
             else:
                 original_logger_warning(msg, *args, **kwargs)
 
-        asyncio.log.logger.warning = logger_warning
+        asyncio.log.logger.warning = logger_warning  # type: ignore
 
         aggregated_exceptions = AggregatedExceptions()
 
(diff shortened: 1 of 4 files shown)


xiaoxianma:
@deathowl Trying to use Click, there are more places to be refactored than I thought. I ended up with making _Config dataclass.

xiaoxianma:
@deathowl Hmm, it seems that mypy failures are not introduced by this commit. Do you have an idea? Thanks

deathowl (approved):

deathowl:
@xiaoxianma  Thank you for your contribution :)

testslide/__init__.py
```diff
@@ -421,7 +421,7 @@ def showwarning(
             else:
                 original_showwarning(message, category, filename, lineno, file, line)
 
-        warnings.showwarning = showwarning
+        warnings.showwarning = showwarning  # type: ignore
```
fornellas, line 424: I'm curious why this ignore was needed now (same for `logger_warning` below). Is this one of the cases we could / should add an exception at `mypy.ini` as `ignore_missing_imports = True`?

deathowl: Nono, it's ```testslide/__init__.py:441: error: Cannot assign to a method```

fornellas: Something's odd. Just tried this on master (8833026):

- Remove these 2 `type: ignore`.
- Pyenv with Python 3.8.5.
- `python -m venv venv`
- `. venv/bin/activate`
- `make install_build_deps`
- `make mypy`

and... no errors reported! Not sure what extra you have at your machine to detect this issue, but I'd love to have it on master, lol... I think it is time start using [tox](https://tox.readthedocs.io/en/latest/) and have a file with all dependency versions frozen, which we update every release.

PR #252, opened 2020-10-17
Add colored stack traces

### LINKED ISSUE #235: Improve coloring of printed stack traces
Currently stack traces are all cyan:

image

Let's improve things:

- File paths with different color
- Function names with different color
- Syntax highlight for code

The code that prints stack traces is [here](https://github.com/facebookincubator/TestSlide/blob/master/testslide/runner.py#L194-L214).

A test can be added similar to [this one](https://github.com/facebookincubator/TestSlide/blob/master/tests/cli_unittest.py#L268) and [this one](https://github.com/facebookincubator/TestSlide/blob/master/tests/cli_unittest.py#L298) for asserting on colors.

CC @karlodwyer

fornellas:
Because this:

image

is easier to read than this:

image

CC @thatch, @karlodwyer

Closes #235

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,6 +4,8 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 [mypy-psutil.*]
 ignore_missing_imports = True
+[mypy-pygments.*]
+ignore_missing_imports = True
 [mypy-testslide.testslide]
 disallow_untyped_calls = True
 disallow_untyped_defs = True
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ setup(
     setup_requires=["setuptools>=38.6.0"],
     install_requires=[
         "psutil>=5.6.7",
+        "Pygments>=2.7.1",
         "typeguard>=2.9.1",
         'dataclasses==0.6; python_version < "3.7"',
     ],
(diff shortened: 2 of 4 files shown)


thatch:
This looks good from the Pygments side -- there doesn't appear to be a test that forces color to exercise this code path though.

Note that if Pygments can't parse the line (without context) it's possible that it will output text on a red background corresponding to `pygments.token.Error` when it doesn't match the known patterns.  Can you try with a `SyntaxError` like the line `!!` and make sure it's not shown as red using the api you have?

fornellas:
@thatch  tkz for the review! So, it does not blow up:

image

I'm updating the PR to catch the colored output.

thatch:
(looks like there's a missing newline there though?)

fornellas:
Where?

thatch:
The missing newline is after the `!!`; was the SyntaxError-generating file itself missing the newline?

fornellas:
I added the !! to break the parser on purpose, and it simply lost the colors, but didn't blew up.

deathowl (approved):

fornellas:
@thatch  I'm merging this now, but if you feel like there are pending things, comment here I can do another PR on top :-D

PR #261, opened 2020-10-28
util: use python3 for run_silent_if_successful

davide125:
TestSlide requires Python 3, so might as well use it here too. This fixes the build on systems that don't have Python 2 installed as `/usr/bin/python`, such as modern Fedora and CentOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ DIST_TAR_GZ = dist/TestSlide-$(shell cat testslide/version).tar.gz
 # Verbose output: make V=1
 V?=0
 ifeq ($(V),0)
-Q := @python util/run_silent_if_successful.py
+Q := @python3 util/run_silent_if_successful.py
 else
 Q :=
 endif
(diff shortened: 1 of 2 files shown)


fornellas (approved):

fornellas:
Python 2 has been officially dead for almost 1 year now... though I understand some distros lag behind on this.

@davide125 TestSlide's official builds run against [specific Python versions](https://github.com/facebookincubator/TestSlide/blob/master/.travis.yml) and I recently put #254 RFC up for start using tox, which should address Python version issues, though it may mean there's some rework on the Fedora package later on. What do you think?

Topic hijack... @deathowl suggestions on where to go with the build? It seems there's always some small issue here and there.. @fabriziocucci pitched poetry today, which sounds interesting.

michel-slm:
@fornellas using tox should not affect the Fedora package that much. Some Fedora internal tools (such as fedpkg) use tox to make sure they work across different Python environments too - https://pagure.io/fedpkg/tree/master

As long as python setup.py build / install / test or some equivalent works fine without using venv then building in Fedora should be fine (builds are done in mock in a throwaway environment, and there is only one default system Python version per release).

PR #262, opened 2020-10-28
downgrade pygments requirement to 2.6.1

davide125:
Unless I'm mistaken, we don't actually _need_ 2.7.1 for this, and downgrading the requirement to 2.6.1 makes it possible to get TestSlide packaged in Fedora 33.


fornellas:
That's reasonable. Thanks for carrying on the Fedora packaging :-D

PR #263, opened 2020-11-09
Skip finished tasks when checking for async leaked tasks

kchatzii:
**What:**
Updated leaked tasks checking to skip finished coroutines, in order to avoid incorrect failures when tasks have been completed

**Why:**
With Python versions that doesnt include this fix (https://github.com/python/cpython/pull/7174), Testslide will raise a LeftOverActiveTasks exception even if all coroutines have been completed successfully.
Example:
` <Task finished coro=<TestAsyncRun.test_catches_leaked_tasks.<locals>.dummy_async_func() done, defined at *> result=None created at *>`

Although the task is **done**, Testslide will **fail**.

This PR skips finished tasks from the async leak testing

**How:**

- Added extra function get_running_tasks to return only non done tasks
- Updated _async_ensure_no_leaked_tasks to use this new function

**Checklist**:

### DIFF
--- a/testslide/__init__.py
+++ b/testslide/__init__.py
@@ -72,6 +72,10 @@ else:
     get_all_tasks = asyncio.all_tasks
 
 
+def get_running_tasks():
+    return [task for task in get_all_tasks() if not task.done()]
+
+
 class LeftOverActiveTasks(BaseException):
     """Risen when unfinished asynchronous tasks are detected."""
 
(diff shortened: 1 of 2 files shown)


deathowl (approved):

deathowl:
Thanks for the PR @kchatzii

testslide/__init__.py
```diff
@@ -72,6 +72,10 @@ def asyncio_run(coro):
     get_all_tasks = asyncio.all_tasks
 
 
+def get_running_tasks():
```
fornellas, line 75: s/get_running_tasks/get_not_done_tasks/g.

PR #270, opened 2020-11-19
Tweak test report

fornellas:
Tweak the test execution report to:

- Show successful, failed, skipped and not executed tests even when it is 0.
  - It is printed in dim white in this case.
  - When there's 1+ examples, it is printed with the same usual color code (green, red, yellow, cyan).
- Improve the message for slow imports, making it more visible and hinting what to do next.
- Add link to documentation: despite being comprehensive, few people refer to it, let's make this better.
- Add nice colored ASCII colored logo, because why not help branding ¯\_(ツ)_/¯

image

### DIFF
--- a/tests/cli_unittest.py
+++ b/tests/cli_unittest.py
@@ -464,7 +464,6 @@ class TestCliDocumentFormatter(FormatterMixin, TestCliBase):
                 "top context\n"
                 "  *focused example: PASS\n"
                 "\n"
-                "Finished 1 example(s) in "
                 # TODO add remaining bits of the output (using regexes)
             )
         )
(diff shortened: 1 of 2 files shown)


fornellas:
@deathowl do you know about this failing build? https://app.circleci.com/pipelines/github/facebookincubator/TestSlide/2/workflows/135eecd8-7fb6-42bf-979b-77d65e7c4c8b/jobs/2

deathowl:
> @deathowl do you know about these 2 build errors?

i think rebase should resolve them

deathowl (approved):

fornellas:
I'll move this logo to the right in the next week or so, so it is not annoying to users.

fornellas:
Moved logo to the right of the screen so it does not annoy:

- If stdout is a terminal:
  - if columns > 80 then it caps the logo to the 80th column (so it does not goes all the way to the right on big terminals).
  - else
    - if it can fit the logo then print it
    - else does not print it
- else
  - print the logo as if the terminal had 80 columns.

image

PR #271, opened 2021-01-02
move dependency management to requirements.txt and requirements-dev.txt.

deathowl:
Add dockerfile and Docker related make targets

**What:**

- Move dependency management to requirements.txt and requirements-dev.txt to ease local development 
- Add related make targets: make install_build_deps and make install_deps
- Add Dockerfile 
- Add Docker related make targets to build, one-shot test built image, run container with bind-mounted sources in the background, enter said container:
Sample Docker run: https://gist.github.com/deathowl/e30bb72a791e362b3a8cf0e45e56c4f3
**Why:**

To ease development :) 
**How:**

**Risks:**
None
**Checklist**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+# set base image (host OS)
+FROM python:3.8
+
+# set the working directory in the container
+WORKDIR /code
+
+# install make 
+RUN apt-get update && apt-get install -y \
+  make \
+  && rm -rf /var/lib/apt/lists/*
+# copy the dependencies file to the working directory
+COPY requirements.txt .
+# copy the dev-dependencies file to the working directory
+
+COPY requirements-dev.txt .
+
+# install dependencies
+RUN pip install -r requirements.txt
+# install dev dependencies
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
+# set base image (host OS)
+FROM python:3.8
```
fornellas, line 2: 3.9 (latest)?

fornellas:
Absolutely 100% with having a container for development, awesome!

How do you see we maturing it? Right now requirements.txt are "as is", meaning we don't care about versions for most stuff. I thought of eventually updating the GH workflow to auto-update a `pip freeze` snapshot with all latest, so every push we do, we test with all latest versions (another PR).

fornellas (approved):

fornellas:
PS: should we abandon #254 ? (I'm fine if you say so).

deathowl:
@fornellas no, i think tox will be a huge added value, so i definitely wouldn't abandon it.

PR #272, opened 2021-01-04
requirements: relax dataclasses version dependency

davide125:
I don't see anything in TestSlide that would require 0.6 specifically, and this will make it easier to get TestSlide packaged for distributions still stuck on Python 3.6 (like EPEL, where I'm in the process of adding dataclasses 0.8 in https://bugzilla.redhat.com/show_bug.cgi?id=1912143).


deathowl:
Merged. Thanks for the PR :)

PR #273, opened 2021-01-04
requirements: allow Pygments 2.2.0 or later

davide125:
Trying to get TestSlide to build for EPEL in https://bugzilla.redhat.com/show_bug.cgi?id=1912139 and one issue is that RHEL 8 / CentOS 8 only ships with Pygments 2.2.0. I *think* it should work just fine, as I don't see anything in the changelog between 2.2.0 and 2.6.1 that would affect TestSlide directly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 psutil>=5.6.7
-Pygments>=2.6.1
+Pygments>=2.2.0
 typeguard>=2.10.0
 dataclasses>=0.6; python_version < "3.7"


fornellas:
Ahh.... old RH distros... #262 wasn't enough.

#252 added a test for this, but our build system (see GH actions) will run with latest pygments. @davide125 do you mind doing a pyenv and confirming the build (`make`) passes with this outdated pygments version please? I have a faint memory of @thatch mentioning we **may** depend on newer versions (I could be wrong though).

davide125:
Looks like all tests are passing with 2.2.0 in a virtualenv for me.

davide125:
Full output: https://gist.github.com/davide125/e7ccb64630cd85c3f43425e5bac4345f

thatch:
The specific lexer and formatter you're using have existed for years; I think you're fine to declare `pygments>=2.2.0` based on a quick visual inspection.

fornellas:
Downgrade it is!

PR #285, opened 2021-02-21
Ensure that type validation is executed for StrictMock instances as well

macisamuele:
While debugging some issue into some FB internal tests (which I was expecting failing) I've noticed that using `mock_callable` / `mock_async_callable` on `StrictMock` does not enable type validation.
This feels like an oversight on the code and has the potential to have users (myself included) writing tests expecting the framework to be validating the types (and as such ensuring that the mocks are consistent with the typing annotations).

A super small representative example is available on https://gist.github.com/macisamuele/b6e85ddfad7f5802ebedd5a2a0810c6c

This PR does also fixes/improves few more details:
* ensure that functions wrapped for type checking are not altering function name and docstring (debuggability)
* removes redundant `FORCE` Makefile target
* updates the repository link in `setup.py`

## Why:

These changes are needed to ensure that `mock_callable` and `mock_async_callable` does preserve type validation regardless of the target being a *real* instance or a `StrictMock` instance.

**NOTE**: In order to make the change possible and well defined it does modify the `tests.sample_module` (6be4622482e77b45ce240ffbf8804ac75fbb30ae) in order to have the methods returning `Iterable`s instead of `str`s.
This is needed because current testing (`mock_callable_testslide.py`) does mock `class_method` with `to_yield_values` which is incompatible from a typing prospective. Fixing the tests would have required more work respect fixing the module class and it is safe because the module is used only for testing purposes.

## How:

The `mock_callable` implementation was intentionally excluding `StrictMock` instances to be wrapped by type validation, this PR *just* removed this exclusion.

## Risks:

Users of the library which have been relying on the lack of type validation on mocked functions on `StrictMock`s would see their tests failing.
This is intentional as the lack of the type validation in such cases was a bug and not considered a feature.
For these users, my suggestion would be to fix their tests or to add `type_validation=False` into the `mock_(async_)callable` definition (the later is discouraged as it hides the issue instead of solving it)

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,9 +36,6 @@ endif
 .PHONY: all
 all: tests coverage_report docs sdist
 
-# .PHONY does not work for implicit rules, so we FORCE them
-FORCE:
-
 ##
 ## Docs
 ##
@@ -57,7 +54,7 @@ docs_clean:
 ## Tests
 ##
 
-%_unittest.py: FORCE coverage_erase
+%_unittest.py: coverage_erase
 	@printf "${TERM_BRIGHT}UNITTEST $@\n${TERM_NONE}"
 	${Q} coverage run \
 		-m unittest \
@@ -70,15 +67,15 @@ unittest_tests: $(TESTS_SRCS)/*_unittest.py
 
 .PHONY: pytest_tests
 pytest_tests: export PYTHONPATH=${CURDIR}/pytest-testslide:${CURDIR}
-pytest_tests: FORCE coverage_erase
+pytest_tests: coverage_erase
 	@printf "${TERM_BRIGHT}INSTALL pytest_testslide DEPS ${TERM_NONE}\n"
 	${Q} pip install -r pytest-testslide/requirements.txt
 	@printf "${TERM_BRIGHT}PYTEST pytest_testslide${TERM_NONE}\n"
 	${Q} coverage run \
-	-m pytest \
-	pytest-testslide/tests
+		-m pytest \
+		pytest-testslide/tests
 
-%_testslide.py: FORCE coverage_erase
+%_testslide.py: coverage_erase
 	@printf "${TERM_BRIGHT}TESTSLIDE $@\n${TERM_NONE}"
 	${Q} coverage run \
 		-m testslide.cli \
--- a/testslide/lib.py
+++ b/testslide/lib.py
@@ -292,7 +292,14 @@ def _wrap_signature_and_type_validation(
     with_sig_and_type_validation.__qualname__ = "TestSldeValidation({})".format(
         with_sig_and_type_validation.__qualname__
     )
+    setattr(  # noqa: B010
+        with_sig_and_type_validation, "__is_testslide_type_validation_wrapping", True
+    )
     return with_sig_and_type_validation
+
+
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -36,9 +36,6 @@ endif
 .PHONY: all
 all: tests coverage_report docs sdist
 
-# .PHONY does not work for implicit rules, so we FORCE them
-FORCE:
-
```
fornellas, line 41: Things are indeed worknig now, thanks for cleaning this up!

testslide/lib.py
```diff
@@ -271,6 +272,8 @@ def _wrap_signature_and_type_validation(
 
     skip_first_arg = _skip_first_arg(template, attr_name)
 
+    # Add this so docstrings and method name are not altered by the mock
+    @wraps(callable_template)
```
fornellas, line 276: Hum... I do understand the idea here, though I'm weighting if letting it explicit "this is a mock", with new name and docstring would be better than "implicit" and keeping the old names... Not strongly opinionated either way though.

fornellas:
LGTM \o/ Thanks for fixing this totally overseen case!

testslide/lib.py
macisamuele: The idea here is that I would like to prevent as much as possible to have
```
<function _wrap_signature_and_type_validation.<locals>.with_sig_and_type_validation at 0x10ee343b0>
```
as information of the wrapped function.

Doing so makes debugging very very hard (at least I had a particularly hard time to identify the mocked function in `mock_callable_testslide.py` due to the function name).

With this change the function would look like
```
<function Client.sync_function at 0x102386ef0>
```

I can see your point of preserving the information of "this is a mock", in this case we might want to have something like
```
<function TestSldeValidation(_wrap_signature_and_type_validation) at 0x105c17560>
```

If you might like this idea I can update the PR

macisamuele:
I've rebased the branch on top of the latest master and addressed the suggestion about preserving "this is a mock" concept and removed a miscellaneous `facebookincubator` links

Other than that the new iteration changes the return types of `sample_module` to `List[str]` from `Iterable[str]` mostly because we would risk to have inconsistent types due to the fact that `"something"` is either a `str` and an `Iterable[str]` and this might cause confusion on tests (especially if you are expecting them failing and they are not).

Last but not least I've improved the handling of `type_validation` into `mock_callable` and `mock_async_callable`.
Before the changes (in 026af9d) we would default `type_validation` to `True` fully ignoring that the underlying `StrictMock` (if present) is defining it to `None` and the same was used to happen on the other direction where we would set `type_validation=False` into `mock_callable` definition but then type validation would still be performed.
**NOTE**: I faced this while fixing some tests internally to FB codebase and it felt interesting to fix-forward in the OS as well. The provided tests (in `tests/strict_mock_testslide.py`) highlight the issue.
In order to fix this I've decided to relax `type_validation` from a boolean to a ternary value where
* `None` means that type validation would be enable (except if the underlying strictmock would instruct otherwise)
* `True` will ensure that it is enabled
* `False` will ensure that it is disabled

I've not forgotten about this diff, but we uncovered a lot test cases breaking internally because of the new detection :)

deathowl (approved):

PR #287, opened 2021-02-24
Fix master

fornellas:
The tests were correctly breaking after a clowny change in master (guilty!). This PR fixes the object used in the test, so it passes.


macisamuele (approved):
Green tests is all it takes :)

Issue #233, opened 2020-09-01
Add constructor input checks for matchers


fornellas:
When we do:

```
AnyInstanceOf(not_a_type)
```

we end up raising an exception **not** at the constructor, but rather when the comparison is being made, and `isinstanceof` (at `AnyInstanceOf`'s code) will complain that `not_a_type` is... not a type.

Let's revisit all matchers & ensure that all of them raise `ValueError` if the constructor arguments are invalid from the constructor itself.

CC @deathowl

PR #289, opened 2021-03-16
raising ValueError at the constructor

EricLiclair:
**What:**

https://github.com/facebook/TestSlide/issues/233

**Why:**

An error was raised when the comparison of two non-similar type data were compared. To be more technically correct, the error should be raised at the time of creating an instance. And thereby this solution.

**How:**

Compared the input arguments, passed at the time of constructor call, with the type of class Matcher. If the two type do not match, a Value Error is raised.

**Risks:**

No Such risks are possible as far as I understand the use of the particular constructors.

**Checklist**:

- [ ] Added tests, if you've added code that should be tested N/A
- [ ] Updated the documentation, if you've changed APIs N/A

I need some opinion about what exactly to be written for the error message.

### DIFF
--- a/testslide/matchers.py
+++ b/testslide/matchers.py
@@ -53,6 +53,8 @@ class _AndMatcher(_AlreadyChainedMatcher):
     """
 
     def __init__(self, a: Matcher, b: Matcher) -> None:
+        if type(a) != type(Matcher()) or type(b) != type(Matcher()):
+            raise ValueError("Not of type Matcher")
         self.a = a
         self.b = b
 
@@ -65,6 +67,8 @@ class _AndMatcher(_AlreadyChainedMatcher):
 
 class _XorMatcher(_AlreadyChainedMatcher):
     def __init__(self, a: Matcher, b: Matcher) -> None:
+        if type(a) != type(Matcher()) or type(b) != type(Matcher()):
+            raise ValueError("Not of type Matcher")
         self.a = a
         self.b = b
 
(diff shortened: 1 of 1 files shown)


testslide/matchers.py
```diff
@@ -53,6 +53,8 @@ class _AndMatcher(_AlreadyChainedMatcher):
     """
 
     def __init__(self, a: Matcher, b: Matcher) -> None:
+        if type(a) != type(Matcher()) or type(b) != type(Matcher()):
```
macisamuele, line 56: I'm wondering if we should be testing types via `isinstance` instead of exact type as in here

```suggestion
        if not isinstance(a, Matcher) or not isinstance(b, Matcher)):
```

The same would apply to the other cases as well

NOTE: I haven't yet checked the implications of the changes.

EricLiclair: Thanks for your reply, I'll get it done. Right now now I am facing some issues setting up the development environment. The moment I get it done. I'll implement the suggested changes. 😄

EricLiclair: Done! 😄
@macisamuele 

I added a [new PR](https://github.com/facebook/TestSlide/pull/290) after resolving the installation and environment issues.

macisamuele: We're working on making it better and streamlined.
I would suggest to create a virtual environment, activate it and the run `make install_build_deps install_deps` so all the dependencies will be available there

EricLiclair: I did the same. But it threw an error stating that those modules were not found. I had to explicitly install `typeguard` and `psutil` to make it work.

PR #291, opened 2021-03-22
Raise Exception from AnyInstanceOf Constructor

EricLiclair:
**What:**

#233 

**Why:**

**How:**

Checked for the input arguments of the `AnyInstanceOf` to be of an instance of `type`. If not, raise a `ValueError` exception.

**Risks:**

**Checklist**:

Do I need to add the "correct usage" of the AnyInstanceOf in the tests?

### DIFF
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -134,12 +134,8 @@ class GenericTestCase(testslide.TestCase):
         self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
 
     def testTriggeringAnyInstanceOfException(self):
-        c = sample_module.SomeClass()
         with self.assertRaises(ValueError):
-            self.mock_callable(c, "method").for_call(
-                testslide.matchers.AnyInstanceOf(2)
-            ).to_return_value("mocked_response")
-            c.method()
+            testslide.matchers.AnyInstanceOf(2)
 
     def testAnyWithCall(self):
         self.assertEqual(testslide.matchers.AnyWithCall(lambda x: "b" in x), "abc")
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -131,8 +131,6 @@ class GenericTestCase(testslide.TestCase):
 
     def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
-
-    def testTriggeringAnyInstanceOfException(self):
         self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
         with self.assertRaises(ValueError):
             testslide.matchers.AnyInstanceOf(2)
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -131,9 +131,9 @@ class GenericTestCase(testslide.TestCase):
 
     def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
-        self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
 
     def testTriggeringAnyInstanceOfException(self):
+        self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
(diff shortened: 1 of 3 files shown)


macisamuele (approved):
@EricLiclair thanks for contributing

> Do I need to add the "correct usage" of the AnyInstanceOf in the tests?

`testAnyInstanceOf` does already check for it ;)

tests/sample_module.py
```diff
@@ -19,7 +19,7 @@ class SomeClass:
     other_class_attribute = OtherClass
     typedattr: str = "bruh"
 
-    def method(self):
+    def method(self, *args, **kwargs):
```
EricLiclair, line 22: This was used for writing a different test case which I removed before committing. The removed test was to test the "correct usage" of the `AnyInstanceOf `using `mock_callable`. If asked to, I'll add that test again.

EricLiclair:
Thank you @macisamuele for your help. Learned a lot in the past week. 😊

tests/sample_module.py
macisamuele, line 22: If these changes are no longer needed then they can be removed

tests/matchers_unittest.py
```diff
@@ -133,6 +133,14 @@ def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
         self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
 
+    def testTriggeringAnyInstanceOfException(self):
+        c = sample_module.SomeClass()
+        with self.assertRaises(ValueError):
+            self.mock_callable(c, "method").for_call(
+                testslide.matchers.AnyInstanceOf(2)
+            ).to_return_value("mocked_response")
+            c.method()
```
macisamuele, line 142: `testslide.matchers.AnyInstanceOf(2)` would raise the exception so invoking `c.method()` is not needed as well as creating the `c` instance is not neede.

Yu might want to have this test more inline with the other tests

```suggestion
        self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
        with self.assertRaises(ValueError):
            testslide.matchers.AnyInstanceOf(2)
```

tests/sample_module.py
EricLiclair: Shall I remove it and commit again?

tests/matchers_unittest.py
```diff
@@ -131,7 +131,11 @@ def testAny(self):
 
     def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
+
+    def testTriggeringAnyInstanceOfException(self):
```
macisamuele, line 135: nit. I would suggest to keep a single test just to keep it uniform with the rest of the tests

macisamuele (commented):
I know that I might asking a lot, so sorry in advance.
Would you have bandwidth to address also the other matches? (possibility on a different PR such that this can eventually be merged)

tests/matchers_unittest.py
EricLiclair: Okay, I believe the `self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)` was unintentionally selected to be added to the triggering test.

I'll keep it where it was initially, in `testAnyInstanceOf`.
And only add a single triggerTest to `testTriggeringAnyInstanceOfException`.

Is it what I am expected to do?

EricLiclair:
> I know that I might asking a lot, so sorry in advance.
> Would you have bandwidth to address also the other matches? (possibility on a different PR such that this can eventually be merged)

And yes, no worries of the bandwidth.
Also, I might sometimes ask childish questions 😅. Just to be sure about what needs to be done. I hope it won't be an issue.

macisamuele:
> And yes, no worries of the bandwidth.
> Also, I might sometimes ask childish questions 😅. Just to be sure about what needs to be done. I hope it won't be an issue.

at this point I do believe that you have fully grasped what needs to be done and how to do it. Still, no questions are childish so do not hesitate to post them.

PR #292, opened 2021-03-24
This PR aims to fix Issue #233

EricLiclair:
**What:**

Fix for issue #233 
1. Added Constructor level input type checks for various Matchers. 
2. Added missing tests for `testAnyFloat`, `testNotEmptyList` and `testEmptyList`

**Why:**

**How:**

_For point 1_
Checked for the passed arguments to be of the specified types. If not, raised a `ValueError` with appropriate error message.

**Risks:**

N/A. At least not intentionally.

**Checklist**:

### DIFF
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -134,12 +134,8 @@ class GenericTestCase(testslide.TestCase):
         self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
 
     def testTriggeringAnyInstanceOfException(self):
-        c = sample_module.SomeClass()
         with self.assertRaises(ValueError):
-            self.mock_callable(c, "method").for_call(
-                testslide.matchers.AnyInstanceOf(2)
-            ).to_return_value("mocked_response")
-            c.method()
+            testslide.matchers.AnyInstanceOf(2)
 
     def testAnyWithCall(self):
         self.assertEqual(testslide.matchers.AnyWithCall(lambda x: "b" in x), "abc")
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -131,8 +131,6 @@ class GenericTestCase(testslide.TestCase):
 
     def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
-
-    def testTriggeringAnyInstanceOfException(self):
         self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
         with self.assertRaises(ValueError):
             testslide.matchers.AnyInstanceOf(2)
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -131,9 +131,9 @@ class GenericTestCase(testslide.TestCase):
 
     def testAnyInstanceOf(self):
         self.assertEqual(testslide.matchers.AnyInstanceOf(str), "durrdurr")
-        self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
 
     def testTriggeringAnyInstanceOfException(self):
+        self.assertNotEqual(testslide.matchers.AnyInstanceOf(str), 7)
         with self.assertRaises(ValueError):
             testslide.matchers.AnyInstanceOf(2)
 
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -208,10 +208,6 @@ class StringTest(testslide.TestCase):
         self.assertEqual(testslide.matchers.RegexMatches("b[aeiou]t"), "bott")
         self.assertNotEqual(testslide.matchers.RegexMatches("b[aeiou]t"), "boot")
         self.assertNotEqual(testslide.matchers.RegexMatches("b[aeiou]t"), 13)
-        with self.assertRaises(ValueError):
-            testslide.matchers.RegexMatches("b[aeiou]t", 12.0)
-        with self.assertRaises(ValueError):
-            testslide.matchers.RegexMatches(1234, 12.0)
 
     def testStrContaining(self):
         self.assertEqual(testslide.matchers.StrContaining("bot"), "bott")
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -27,6 +27,7 @@ class IntMatcherTest(testslide.TestCase):
         self.assertEqual(testslide.matchers.IntBetween(21, 666), 42)
         self.assertEqual(testslide.matchers.IntBetween(21, 666), 21)
         self.assertEqual(testslide.matchers.IntBetween(21, 666), 666)
+        self.assertEqual(testslide.matchers.IntBetween(21.0, 666.0), 666)
         self.assertNotEqual(testslide.matchers.IntBetween(42, 69), 666)
         self.assertNotEqual(testslide.matchers.IntBetween(42, 69), "derp")
         with self.assertRaises(ValueError):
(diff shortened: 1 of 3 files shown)


tests/matchers_unittest.py
```diff
@@ -18,75 +18,132 @@ def test_NotThisInt(self):
...
+        with self.assertRaises(ValueError):
+            testslide.matchers.NotThisFloat(10)
 
     def test_FloatBetween(self):
-        self.assertEqual(testslide.matchers.FloatBetween(2.1, 666), 4.2)
-        self.assertEqual(testslide.matchers.FloatBetween(2.1, 666), 2.1)
-        self.assertEqual(testslide.matchers.FloatBetween(21, 66.6), 66.6)
+        self.assertEqual(testslide.matchers.FloatBetween(2.1, 666.0), 4.2)
+        self.assertEqual(testslide.matchers.FloatBetween(2.1, 666.0), 2.1)
+        self.assertEqual(testslide.matchers.FloatBetween(21.0, 66.6), 66.6)
```
EricLiclair, line 98: This is where the tests were updated as the argument used `666` was not of type `float`.

tests/matchers_unittest.py
```diff
@@ -175,6 +248,20 @@ def testAnyList(self):
         self.assertNotEqual(testslide.matchers.AnyList(), 69)
         self.assertNotEqual(testslide.matchers.AnyList(), None)
 
+    def testNotEmptyList(self):
```
EricLiclair, line 251: Added tests for `notEmptyList` followed by tests for `emptyList`

tests/matchers_unittest.py
```diff
@@ -18,75 +18,132 @@ def test_NotThisInt(self):
...
         self.assertNotEqual(testslide.matchers.IntLessOrEquals(21), 22)
         self.assertNotEqual(testslide.matchers.IntLessOrEquals(42), "derp")
+        with self.assertRaises(ValueError):
+            testslide.matchers.IntLessOrEquals("derp")
+        with self.assertRaises(ValueError):
+            testslide.matchers.IntLessOrEquals(42.42)
 
 
 class FloatMatcherTest(testslide.TestCase):
+    def testAnyFloat(self):
```
EricLiclair, line 81: Added tests for `anyFloat`

deathowl:
@EricLiclair  build seems to break for Python3.9

EricLiclair:
@deathowl I did notice the failed build test, but i am not sure why is that happening. I checked the test log and found that `tests/mock_async_callable_testslide.py` is where the test fails but I have not changed anything in that file. Can you suggest me something to get this fixed?

Also, the make instruction successfully executes in my system. I wonder where am I going wrong.


I also forked the master repo from my second account just to be sure I was not missing anything, added my changes and dropped a PR. Seems the build failed for that PR as well. I am kind of stuck. I'll be looking into this, but your comments would be appreciated.

tests/matchers_unittest.py
```diff
@@ -18,75 +18,132 @@ def test_NotThisInt(self):
...
+        self.assertEqual(testslide.matchers.AnyFloat(), 666.0)
+        self.assertEqual(testslide.matchers.AnyFloat(), 4.2)
+        self.assertNotEqual(testslide.matchers.AnyFloat(), "derp")
+
```
EricLiclair, line 85: Added tests for `anyFloat`

tests/matchers_unittest.py
EricLiclair, line 98: This is where the tests were updated as the argument used `666` was not of type `float`.

tests/matchers_unittest.py
```diff
@@ -175,6 +244,20 @@ def testAnyList(self):
...
+        self.assertNotEqual(testslide.matchers.NotEmptyList(), {})
+        self.assertNotEqual(testslide.matchers.NotEmptyList(), None)
+
+    def testEmptyList(self):
+        self.assertEqual(testslide.matchers.EmptyList(), [])
+        self.assertNotEqual(testslide.matchers.EmptyList(), ["a", "b"])
+        self.assertNotEqual(testslide.matchers.EmptyList(), 69)
+        self.assertNotEqual(testslide.matchers.EmptyList(), {})
+        self.assertNotEqual(testslide.matchers.EmptyList(), None)
+
```
EricLiclair, line 260: Added tests for `notEmptyList` followed by tests for `emptyList`

testslide/matchers.py
```diff
@@ -265,7 +313,7 @@ def __init__(self) -> None:
 
 class RegexMatches(Matcher):
     """
-    Compares true if other mathes given regex.
+    Compares true if other matches given regex.
     """
 
     def __init__(self, pattern: str, flags: int = 0) -> None:
```
EricLiclair, line 319: **What I would like to mention is that, adding a type check here was the reason the build was failing. I wonder why was it so. And hence I couldn't add a type check here. Any comments on this will be appreciated.**

EricLiclair:
@deathowl hey, sorry that it took a lot of time figuring out what went wrong.

Also, @macisamuele I believe this should work. ☺
Review awaited.

tests/matchers_unittest.py
macisamuele: I need to go through all the PR to fully review it.

a suggestion that I would provide, by checking this test, I'd that it would consider acceptable to have integers in the bounds of floatbeetween,after all the intention is to get numbers (floats) within 2 ranges

most probably the same comment will apply to greater and lesser matches as well

testslide/matchers.py
macisamuele: I'm not sure I'm following you ... what failure specifically are you referring to?

Checking the history of your commits I see that the code was something like
```python
def __init__(self, pattern: str, flags: int = 0) -> None:
    if not isinstance(pattern, str) or not isinstance(flags, int):
        raise ValueError(
            f"RegexMatches(...) expects 'str' and 'int' as arguments while '{type(pattern).__name__}' and '{type(flags).__name__}' were provided"
        )
    ...
```
how is this causing random test failures?

---

Side (and personal) note: to make the code-review easier I would suggest to limit merging master branch in your branch.
This because it makes harder to trace changes and makes the history more confusing. Clearly, this does not mean that you should not be doing it but do it parsimoniously.
I do tend to use `git rebase` on the PR branch if the PR takes a very long time to be reviewed and merged in order to gather all the updates.
If a PR is short-living it would be unlikely to have failures caused by other changes and anyway when github actions runs tests they are run on top of your branch merged with master, so tests will detect if your changes would cause test failures caused by new code.

testslide/matchers.py
```diff
@@ -191,31 +191,55 @@ def __init__(self) -> None:
...
+        if not isinstance(ne, int):
+            raise ValueError(
+                f"NotThisInt(...) expects an 'int' as argument while '{type(ne).__name__}' was provided"
+            )
         super().__init__(ne=ne)
 
 
 class IntBetween(_IntComparison):
     def __init__(self, lower: int, upper: int) -> None:
+        if not isinstance(lower, int) or not isinstance(upper, int):
```
macisamuele, line 203: I would personally allow float values as well.
I mean what if I'm looking for any integer between 2.8 and 33.9 ?

Think about a case like
```python
def test_dummy(self) -> None:
    lower = some_calculation()  # This returns a number, maybe an integer value but probably stored in float
    upper = lower * 2
    IntBetween(lower, upper)  # This would fail, but from a conceptual point of view it might be valid
```

This comment is valid for all the Int* matchers

testslide/matchers.py
```diff
@@ -227,31 +251,55 @@ def __init__(self) -> None:
...
+        if not isinstance(ne, float):
+            raise ValueError(
+                f"NotThisFloat(...) expects a 'float' as argument while '{type(ne).__name__}' was provided"
+            )
         super().__init__(ne=ne)
 
 
 class FloatBetween(_FloatComparison):
     def __init__(self, lower: float, upper: float) -> None:
+        if not isinstance(lower, float) or not isinstance(upper, float):
```
macisamuele, line 263: I would allow integers values as well.
The same reasoning of `IntBetween` holds in this case.

Let's say that I want to match any number that is between 0 and 100 ... why should I be forced to represent them as floats?

tests/matchers_unittest.py
```diff
@@ -18,75 +18,132 @@ def test_NotThisInt(self):
...
 
     def test_IntGreaterThan(self):
         self.assertEqual(testslide.matchers.IntGreaterThan(21), 42)
         self.assertNotEqual(testslide.matchers.IntGreaterThan(21), 21)
         self.assertNotEqual(testslide.matchers.IntGreaterThan(21), 20)
         self.assertNotEqual(testslide.matchers.IntGreaterThan(42), "derp")
+        with self.assertRaises(ValueError):
+            testslide.matchers.IntGreaterThan("derp")
+        with self.assertRaises(ValueError):
+            testslide.matchers.IntGreaterThan(42.42)
```
macisamuele, line 47: I would not make it fail in this case ...

The "concept" here is to have all the integer bigger than `42.42` so numbers like 43, 44, etc. I would say that constraining floats away from here would require users of the matchers to make more gymnastics with `floor`/`ceil`/`int` which might end up with tests that would be less obvious to be read

macisamuele (commented):
@EricLiclair  thanks a lot for your PR.
I do like the general approach used, I would just relax a little bit the constraints here in order to not have a huge toll on developers using it.

I would still be curious to understand what was causing the test failures. Is this something that we can address in this PR such that we can re-add the `ValueError` throwing into `RegexMatches` as well?

macisamuele:
@deathowl : do you have strong opinion against ensuring that we still run full tests on all the python versions even if a single version failed?
I do find this information useful to triage failures which might affect one python version but not another.
This should be rather a simple update on the workflow, just wondering if you have adverse opinions.

testslide/matchers.py
EricLiclair: I wanted to point the PR checks. 
as mentioned by `deathowl` [here](https://github.com/facebook/TestSlide/pull/292#issuecomment-806680047)

I was curious why the build was failing so i checked for the build logs as i mentioned [here](https://github.com/facebook/TestSlide/pull/292#issuecomment-806760133)

[This](https://github.com/facebook/TestSlide/pull/292/checks?check_run_id=2198950761) is the failure i was mentioning about.

And, I'll remember the note.

EricLiclair: In fact all of the commits were failing build test, until the last one.

tests/matchers_unittest.py
EricLiclair: Insightful, I really missed out the point that both `float` and `integer` values can be valid range extremes. I'll try to update accordingly as per the comments you mentioned. Thanks for your review. ☺

deathowl:
@macisamuele  i'll resurrect Fabios Tox diff, and make it work for running all our tests . I think running against all supported py versions is strong signal

deathowl:
Hey @EricLiclair  build still seems broken to me

EricLiclair:
@deathowl @macisamuele  Sorry for being unavailable for the past few months. Hard times.
I added the fix to this in #303 .
This PR can now be closed. Thank you.

deathowl:
Closing because resubmitted in #303

PR #294, opened 2021-03-25
Added Type checking at constructors for different matchers + Added Mi…

eric-liclair:
**What:**

1. Added Constructor level input type checks for various Matchers. 
2. Updated testcases in `test_FloatBetween` which had wrongly used test argument types. (will be pointed out in the comments)
3. Added missing test cases for `testAnyFloat`, `testNotEmptyList` and `testEmptyList`

**Why:**

_Point no. 2_ (Updating testcases in `test_FloatBetween`)
The updates were necessary because without proper argument types the **tests threw** "expected" _ValueErrors_ which is what this PR aims to solve.

**How:**

_For point 1_
Checked for the passed arguments to be of the specified types. If not, raised a `ValueError` with appropriate error message.

_For point 2_
Simply added `.0` at the end of the integer arguments passed so as to explicitly convert them into floats.
This ensures that the nature of the tests were not changed and the expected behavior of any particular test remained the same.

**Risks:**

N/A. At least not intentionally.

**Checklist**:

This PR is also for testing if there is something that am doing wrong.

### DIFF
--- a/tests/matchers_unittest.py
+++ b/tests/matchers_unittest.py
@@ -18,6 +18,10 @@ class IntMatcherTest(testslide.TestCase):
         self.assertEqual(testslide.matchers.NotThisInt(666), 42)
         self.assertNotEqual(testslide.matchers.NotThisInt(69), 69)
         self.assertNotEqual(testslide.matchers.NotThisInt(42), "derp")
+        with self.assertRaises(ValueError):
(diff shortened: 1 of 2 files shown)


eric-liclair:
kindly --- Ignore this PR

Thank you

PR #297, opened 2021-04-07
Use get_typing_hints instead of __annotations__ to resolve types in Python 3.10

### LINKED ISSUE #296: patch_attribute issue with Python 3.10
TestSlide 2.6.4 fails with Python 3.10.0a6.

Failures:
```
  1) patch_attribute(), when target is a StrictMock, with a template, type validation: it fails if new value is of incompatible type
    1) AssertionError: TypeCheckError not raised
      File "testslide/runner.py", line 639, in run
        self._run_example(example)
      File "testslide/runner.py", line 623, in _run_example
        _ExampleRunner(example, self.formatter).run()
      File "testslide/__init__.py", line 540, in run
        self._sync_run_all_hooks_and_example(context_data)
      File "testslide/__init__.py", line 512, in _sync_run_all_hooks_and_example
        aggregated_exceptions.raise_correct_exception()
      File "testslide/__init__.py", line 289, in raise_correct_exception
        raise self.exceptions[0]
      File "testslide/__init__.py", line 273, in catch
        yield
      File "testslide/__init__.py", line 503, in _sync_run_all_hooks_and_example
        self._fail_if_coroutine_function(self.example.code, context_data)
      File "testslide/__init__.py", line 471, in _fail_if_coroutine_function
        return func(*args, **kwargs)
      File "tests/patch_attribute_testslide.py", line 119, in it_fails_if_new_value_is_of_incompatible_type
        self.patch_attribute(self.target, "typedattr", 123)
      File "/usr/lib64/python3.10/unittest/case.py", line 226, in __exit__
        self._raiseFailure("{} not raised".format(exc_name))
      File "/usr/lib64/python3.10/unittest/case.py", line 163, in _raiseFailure
        raise self.test_case.failureException(msg)
```
See for logs and more details: https://bugzilla.redhat.com/show_bug.cgi?id=1944109

tirkarthi:
**What:**

Fixes #296 

**Why:**

Due to PEP 563 becoming default in Python 3.10 the `__annotations__` value doesn't store types and it stores string values. Using `typing.get_type_hints` ensures the types are evaluated and returned like behavior before Python 3.10 .

**How:**

**Risks:**

**Checklist**:

- [ ] Added tests, if you've added code that should be tested
- [ ] Ensured the test suite passes
- [ ] Made sure your code lints
- [ ] Completed the Contributor License Agreement ("CLA")

### DIFF
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -7,7 +7,17 @@ import dis
 import inspect
 import os.path
 from types import FrameType
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, get_type_hints
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Type,
+    Union,
+    get_type_hints,
+)
 
 import testslide.lib
 import testslide.mock_callable
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -634,7 +634,11 @@ class StrictMock(object):
             return
 
         if self._template is not None:
-            annotations = get_type_hints(self._template)
+            try:
+                annotations = get_type_hints(self._template)
+            except KeyError:
+                # Some modules can throw KeyError : https://bugs.python.org/issue41515
+                annotations = {}
             if name in annotations:
                 testslide.lib._validate_argument_type(annotations[name], name, value)
 
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -623,9 +623,10 @@ class StrictMock(object):
         ):
             return
 
-        annotations = get_type_hints(self._template)
-        if name in annotations:
-            testslide.lib._validate_argument_type(annotations[name], name, value)
+        if self._template is not None:
+            annotations = get_type_hints(self._template)
+            if name in annotations:
+                testslide.lib._validate_argument_type(annotations[name], name, value)
 
     def __validate_and_wrap_mock_value(self, name: str, value: Any) -> Any:
         if self._template:
--- a/testslide/strict_mock.py
+++ b/testslide/strict_mock.py
@@ -7,7 +7,7 @@ import dis
 import inspect
 import os.path
 from types import FrameType
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union, get_type_hints
 
 import testslide.lib
 import testslide.mock_callable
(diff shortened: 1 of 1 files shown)


deathowl:
Hey @tirkarthi , Thanks for the PR, please ```make format ```, otherwise it looks good, we'll add a Python 3.10 Target to our builds

tirkarthi:
Thanks @deathowl I have fixed the formatting issues. I will wait if someone has issues on using get_type_hints by default to emulate behavior before Python 3.10 and https://www.python.org/dev/peps/pep-0563/#backwards-compatibility

testslide/strict_mock.py
```diff
@@ -623,8 +633,8 @@ def __validate_attribute_type(self, name: str, value: Any) -> None:
         ):
             return
 
-        if hasattr(self._template, "__annotations__"):
-            annotations = self._template.__annotations__
+        if self._template is not None:
+            annotations = get_type_hints(self._template)
```
deathowl, line 637: This does not play nice with pybind11 based modules
```
in get_type_hints
    base_globals = sys.modules[base.__module__].__dict__
KeyError: 'pybind11_builtins'
```

tirkarthi: Thanks, this seems to have caused an issue in sphinx as well where they catch keyerror and have empty annotations. I don't have the right setup but it will be good to know what was the value of annotations using `__annotations__` attribute in earlier versions.

https://github.com/sphinx-doc/sphinx/issues/8084

deathowl: I guess catching key error and returning empty annotations would work well as a workaround for us as well

tirkarthi: Thanks, I will make the changes. On further search seems this is an open issue in CPython where there is a PR open to catch the KeyError and discard it for these type of cases : https://bugs.python.org/issue41515

testslide/strict_mock.py
```diff
@@ -634,7 +634,11 @@ def __validate_attribute_type(self, name: str, value: Any) -> None:
             return
 
         if self._template is not None:
-            annotations = get_type_hints(self._template)
+            try:
+                annotations = get_type_hints(self._template)
+            except KeyError:
```
deathowl, line 639: can we do except Exception just to make sure we dont break tests when we fail to get typehints for any funny modules?

tirkarthi: KeyError seemed more specific to the modules with different `__module__` . Are there any other failures that expose a bug with get_type_hints? I couldn't see the test failures since it's private.

testslide/strict_mock.py
deathowl: https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-class-decorators it can throw NameError on metaclasses

testslide/strict_mock.py
deathowl: https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-class-decorators it can throw NameError on metaclasses

PR #304, opened 2021-06-13
requirements: relax mypy version dependency

krvaibhaw:
I don't see anything in TestSlide that would require 0.782 specifically, and this will make it easier to get TestSlide packaged for distributions still stuck on Python 3.6. Also, the new mypy release includes many new features and bug fixes :
http://mypy-lang.blogspot.com/2021/06/mypy-0900-released.html

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,7 +3,7 @@ coverage
 coveralls
 flake8
 isort~=5.1
-mypy==0.782
+mypy>=0.782
 ipython
 sphinx
 sphinx-autobuild


deathowl:
Hi @krvaibhaw, Thanks for the PR. Technically for packaging only requirements.txt should matter, and requirements-dev.txt is only used during development. Although this PR makes total sense, and I'm willing to merge if you fix the exposed typechecking issues that resulted from the mypy upgrade . The reason we pinned mypy is to make typechecking step in our tests less fragile

deathowl:
Gentle nudge @krvaibhaw,

deathowl:
I've updated to latest mypy, closing

PR #309, opened 2021-07-13
allow ignoring all other args in for_call except those specified

deathowl:
**What:**
Allow asserting only some call arguments as requested 

**Why:**
Feature request

**How:**

Add new argument ``ignore_other_args`` and ``ignore_other_kwargs`` for mock_callable and mock_async_callable
Change get base_runner.can_accept_args to return True if provided args match, if any of the switches provided
**Risks:**

**Checklist**:


fornellas:
Have you considered using `for_partial_call_args` to keep the idiom of `for_call`?

Issue #276, opened 2021-01-12
Add linter to ensure Copyright header on files


fornellas:
Let's have a linter that checks for this header:

```
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.
```

on all files, and fails tests if missing.

PR #311, opened 2021-08-05
Add linter to ensure Copyright header on files

chetanugale:
**What:**

https://github.com/facebook/TestSlide/issues/276

**Checklist**:

- [ ] Added tests, if you've added code that should be tested  "N/A"
- [ ] Updated the documentation, if you've changed APIs  "N/A"
- [ ] Ensured the test suite passes  "N/A"

Following script will check all python files for COPYRIGHT and display error 
eg:
`"Copyright structure missing or incorrect for:  tests/__init__.py"`


tests/copyright_check/config.py
```diff
@@ -0,0 +1,24 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import sys
+import os
+
+
+debug_logs = False
+print_logs = sys.stderr if debug_logs else open("/dev/null", "w")
```
macisamuele, line 12: Using `/dev/null` creates a dependency on Unix-like platform which is not really needed.

Using something like in the suggestion would allow you to not define the implicit dependency

```suggestion
print_logs = sys.stderr if debug_logs else open(os.devnull, "w")
```

tests/copyright_check/config.py
```diff
@@ -0,0 +1,24 @@
...
+
+validator_path = "tests/copyright_check"
+default_sample_file = "copyright_signature.py"
+
+rootdir = os.path.abspath(os.path.dirname(__file__) + "/../../")
+default_check_dir = os.path.join(rootdir, validator_path)
```
macisamuele, line 18: The presence of `/` into the path assumes unix-like platform ... is this really needed?

Have you tried to consider to have something like the suggestion.
It would allow to prevent platform dependency
```suggestion
from pathlib import Path

default_sample_file = "copyright_signature.py" 
rootdir = Path(__file__).parent.parent.absolute()
default_check_dir = rootdir / "tests" / "copyright_check"
```

NOTE: The type of `rootdir` and `default_check_dir` will now be `pathlib.Path` instead of `str`

tests/copyright_check/config.py
```diff
@@ -0,0 +1,24 @@
...
+
+# if want to run specific files, add to this list
+filelist = list()
+
+# If we want to ignore, add directories in ignore_dirs
+ignore_dirs = list()
```
macisamuele, line 24: nit. please use List literals instead of `list()`

```suggestion
# if want to run specific files, add to this list
filelist = []

# If we want to ignore, add directories in ignore_dirs
ignore_dirs = []
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+
+print_logs = config.print_logs
+
+
+def fetch_samplefile():
+    """
+    Method to fetch sample copyright file
+    :return sample
+    """
+    sample = dict()
```
macisamuele, line 24: nit. ditto: use literals

```suggestion
    """
    Method to fetch sample copyright file
    :return sample
    """
    sample = {}
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+    path = os.path.join(config.default_check_dir, config.default_sample_file)
```
macisamuele, line 25: nit. As you're already leveraging pathlib I would suggest
```suggestion
    path = config.default_check_dir / config.default_sample_file
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+    ext = pathlib.Path(config.default_sample_file).suffix.split(".")[1]
+    sample_file = open(path, "r")
+    ref = sample_file.read().splitlines()
+    sample_file.close()
```
macisamuele, line 29: Please use python context manager while opening files, this is important because if some exceptions would happen within the execution the file is guaranteed to be closed at the OS level and it is more pythonic

```python
    with path.open() as f:
        ref = f.read().splitlines()
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
macisamuele, line 1: The library supports only Python3, I'm wondering if you could provide typing annotations for the defined functions

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+
+import difflib
+
+import os
+import re
+import sys
+import pathlib
+import config
+
+print_logs = config.print_logs
```
macisamuele, line 16: I'm intrigued by the fact that we use the builtin print method to maybe print the logs, and we achieve it by writing into a sink file descriptor.

I'm wondering if having something like

```python
def print_logs(*args, **kwargs):
    if not config.debug_logs:
        return
    print(*args, **kwargs)
```

would be more effective

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+    :return bool
+    """
+    try:
+        f = open(filename, "r")
+    except Exception as exc:
+        print("Unable to open %s: %s" % (filename, exc), file=print_logs)
+        return False
+
+    data = f.read()
+    f.close()
```
macisamuele, line 50: nit: as before please use context managers to deal with files

```suggestion
    try:
        with open(filename, "r") as f:
            data = f.read()
    except Exception as exc:
        print("Unable to open %s: %s" % (filename, exc), file=print_logs)
        return False
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+        ext = get_extension(fpath)
+        basename = os.path.basename(fpath)
+        if ext in extensions or basename in extensions:
+            outfiles.append(fpath)
+    return outfiles
+
+
+def main():
+    formula = dict()
+    formula["py_files"] = re.compile(r"^(#!.*\n)\n*", re.MULTILINE)
```
macisamuele, line 131: nit. use literals
```suggestion
    formula = {"py_files": re.compile(r"^(#!.*\n)\n*", re.MULTILINE)}
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+        (data, _) = p.subn("", data, 1)
+
+    data = data.splitlines()
+
+    # if our test file is smaller than the reference it surely fails!
+    if len(demo) > len(data):
+        print(
+            "File %s smaller than sample file(%d < %d)"
+            % (filename, len(data), len(demo)),
+            file=print_logs,
```
macisamuele, line 71: nit. Any reason to not use f-strings or str.format?

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+        return False
+
+    # trim our file to the same number of lines as the reference file
+    data = data[: len(demo)]
+
+    if demo != data:
+        print(
+            "Header in %s does not match sample file, diff:" % filename, file=print_logs
```
macisamuele, line 80: ditto

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def fetch_files(extensions):
+    """
+    Method to traverse through repo for files
+    :param extensions
+    :return outfiles
+    """
+
+    allfiles = list()
```
macisamuele, line 106: nit. Use literals
```suggestion
    allfiles = list()
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+        for root, dirs, traverse in os.walk(config.rootdir):
+            for d in config.ignore_dirs:
+                if d in dirs:
+                    dirs.remove(d)
+
+            for fname in traverse:
+                fpath = os.path.join(root, fname)
+                allfiles.append(fpath)
+
+    outfiles = list()
```
macisamuele, line 120: nit. Use liteerals
```suggestion
    outfiles = list()
```

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+    for fpath in allfiles:
+        ext = get_extension(fpath)
+        basename = os.path.basename(fpath)
+        if ext in extensions or basename in extensions:
```
macisamuele, line 124: I'm not sure what case you're trying to cover with `basename in extensions`
Can you provide some context?

tests/copyright_check/copyright_validator.py
```diff
@@ -0,0 +1,148 @@
...
+def fetch_files(extensions):
+    """
+    Method to traverse through repo for files
+    :param extensions
+    :return outfiles
+    """
+
+    allfiles = list()
+
+    if len(config.filelist) > 0:
```
macisamuele, line 108: In python an empty list (or container in general) is considered `False` so checking that the len is not 0 or checking if it is `True` or `False` is equivalent.
Moreover, the later approach is more pythonic (commonly present in python codebases)
```suggestion
    if config.filelist:
```

macisamuele (commented):
@chetanugale   thanks for contributing to the project.

In the PR I've left a miscellaneous of comments mostly related to python style and some request of clarification.

One aspect that I'd like to highlight though is that the added `copyright_validator` is not linked to any test run or within github workflows which means that except if a human remember to run it automated checks would not report copyright issues.


In general I would advice to avoid having an "extra-binary" that does check that all the files starts with a given header but rather having a unittest test which roughly does what your binary does.
This because it will be extra easy to integrate with already existing testing platform and it would be easy to have high signals in case of copyright lack

deathowl:
@chetanugale  thanks for the improvements. Py3.10-dev tests failing are not a blocker. Will fix that myself (not related to your chagnes)

macisamuele (approved):
Thanks for applying the feedbacks, left one (but it is not blocking).

Approving this.

PR #333, opened 2022-04-19
Handle type validation for string type hints

williamlw999-fb:
With `from __future__ import annotations`, all type hints become strings. The current type validator assumes that what is being checked is a type and not a string, so no types are being validated in these cases.

The fix is to use Python's typing.get_type_hints to evaluate the type hints at runtime

- [ ] Added tests, if you've added code that should be tested
- [ ] Ensured the test suite passes
- [ ] Made sure your code lints
- [ ] Completed the Contributor License Agreement ("CLA")


deathowl:
LGTM :+1:

fried:
So this pull just needs to run isort? 

Snippet from the failed Checks
```
Formatting errors found, try running 'make format'.
make: *** [Makefile:114: isort] Error 1
Error: Process completed with exit code 2.

```

deathowl:
just run make format

PR #335, opened 2022-07-22
Speed up mock_constructor by removing gc.collect

deathowl:
Summary:
mock_aiog
0:00:00.064367
from this calling gc.collect took:
0:00:00.053297 and it's _synchronous_

since default slow callback detection is 0.1 if someone uses multiple mock_constructor calls these gc.collect-s add up and exceed the threshold

Differential Revision: D38073077

### DIFF
--- a/testslide/mock_constructor.py
+++ b/testslide/mock_constructor.py
@@ -349,7 +349,6 @@ def mock_constructor(
                 "Usage with classes that define __new__() is currently not supported."
             )
 
-        gc.collect()
         instances = [
             obj
             for obj in gc.get_referrers(original_class)


fornellas:
IIRC this GC had to be there for consistency at the check. I belive removing it, will lead to inconsistent corner cases where it'll complain mock will fail, wrongly.

deathowl:
It did not trigger any test faillures on the internal tests,but actually the false slow calllback detection becaause of GC blocking the event loop caused more faulty signals, so the change is reasonable.

fornellas:
I don't think i'd trigger any consistent build issues. IIRC, without this GC, depending on how objects are handled up to that point, there MAY be lingering references the GC would clean up, that, without cleanup, it would then trigger false errors. So, it'd be the kind of unexpected inconsistent behaviour, rather than always failing.

I recommend considering alternatives, such as wrapping this GC run with a temporary disable of the event loop block check.

PR #336, opened 2022-08-05
Remove mock_constructor check for existing instances, as removing gc.collect made it unstable

deathowl:
Summary: Certain weird usecases would require us to call gc.collect(), but that makes the eventloop , so is not a good option. This is a good enough compromise

Differential Revision: D38455488


fornellas:
Agree with the idea. I'd consider adding a big fat warning at the documentation though, regarding the bad cases this check would catch, and let it clear that it is up to users to be mindful of those (and not cut bug reports). Maybe have this check as an opt-in check for mock_constructor?

PR #338, opened 2022-08-09
testslide: add language to sphinx config

davide125:
Summary: This should fix https://bugzilla.redhat.com/show_bug.cgi?id=2113613

Differential Revision: D38540735


davide125:
This was fixed already in https://github.com/facebook/TestSlide/pull/334/commits/5d7dcccd071a5cfbd4b0b93b48e86991b9c17d75 so it's not actually needed.

Issue #346, opened 2023-03-09
Request: Container Matchers


bpjatfb:
The current container matchers only cover `list` and `dict` ([TestSlide docs](https://testslide.readthedocs.io/en/main/patching/argument_matchers/index.html?highlight=matchers#lists)).  Unfortunately, this doesn't help much when I'm working with other containers or iterables (of which, Python has many!).  I am proposing that we add matchers for arbitrary containers as well.  Something like the following, although the names could probably be improved:

- `AnyIterable`: value is iterable
- `AnyIterableContaining`: asserts element exists in container/iterable
- `AnyIterableContainingAll`: asserts container/iterable contains all elements
- `EmptyIterable`: asserts container/iterable is empty

## Example

Actual value (a `dict` of `deque`):
```
{
    'a': deque([1, 2, 3], maxlen=100),
    'b': deque([4, 5, 6], maxlen=100),
    'c': deque([7, 8, 9], maxlen=100),
}
```

Since `maxlen=100` isn't a hard requirement, I don't really want to specify it in my test.  It would be really great if I could test like so:

```
{
    'a': AnyIterableContaining(1, 2, 3),
    'b': AnyIterableContaining(4, 5, 6),
    'c': AnyIterableContaining(7, 8, 9),
}
```

## See also

- Google Test (C++) Container Matchers: https://github.com/google/googletest/blob/main/docs/reference/matchers.md#container-matchers
- Hamcrest (Java/multilanguage): https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains(org.hamcrest.Matcher...)

fornellas:
LPGTM @bpjatfb, how do you feel abotu cutting a PR to add these matches [here](https://github.com/facebook/TestSlide/blob/main/testslide/matchers.py)  and some tests [here](https://github.com/facebook/TestSlide/blob/main/tests/matchers_unittest.py)?

bpjatfb:
Fixed by #350

PR #350, opened 2023-03-27
Collection matchers

bpjatfb:
**What:**

Add Collection matchers, as requested in #346

- ``AnyContaining(element)``: A container that contains ``element``
- ``AnyContainingAll(element_list)``: A container that contains every element of ``element_list``
- ``AnyIterable()``: Any iterable
- ``IterableWithElements(element_list)``: an iterable containing all the elements in ``element_list`` in the same order"
- ``AnyNotEmpty()``: An object where ``len()`` does not evaluate to zero"
- ``AnyEmpty()``: An object where ``len()`` evaluates to zero"

**Why:**

Currently, the official matchers support Lists and Dicts, but not much in the way of abstract collections.  This PR enables patterns like:

```
actual = deque([1, 2, 3], maxlen=100)
self.assertEqual(actual, IterableWithElements([1, 2, 3]))
```

**How:**

I added these matchers to matchers.py alongside their List/Dict counterparts.  I added unittests to matchers_unittest.py. Finally, I added documentation in patching/argument_matchers/.

**Risks:**

No risk of breaking existing code, since these are largely standalone, however there are some edge cases for these matchers that may or may not be intuitive to the user.

- `AnyEmpty`/`AnyNotEmpty` rely on `len()`, so they will throw on `AnyEmpty() == iter([1, 2, 3])`, for example.
- `AnyContaining()`/`AnyContainingAll()` rely on `in`, so will try to match anything that implements `__contains__`
    - They may also exhaust iterators leading to unexpected results.

**Checklist**:

### DIFF
--- a/docs/patching/argument_matchers/index.rst
+++ b/docs/patching/argument_matchers/index.rst
@@ -101,6 +101,18 @@ Dictionaries
 	"``DictContainingKeys(keys_list)``", "A dictionary containing all keys from ``keys_list``"
 	"``DictSupersetOf(this_dict)``", "A dictionary containing all key / value pairs from ``this_dict``"
 
+Containers and Iterables
+------------------------
+.. csv-table::
+	:header: "Matcher", "Description"
+
+	"``AnyContaining(element)``", "Any container that contains ``element``"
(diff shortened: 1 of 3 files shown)


bpjatfb:
Awesome, thanks @deathowl!

PR #359, opened 2024-03-25
Add devcontainer

shish:
Add devcontainer

So that one merely needs to open the project in a supported editor (vscode, github) and the editor will pop up a prompt for "would you like to develop this project inside the container?", and if you click yes then you are straight into a known-good dev environment where you can run "make" and run all the tests

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+{
+	"name": "TestSlide",
+	"build": {
+		"context": "..",
+		"dockerfile": "../Dockerfile"
+	},
+
+	"workspaceMount": "source=${localWorkspaceFolder},target=/root/src/TestSlide,type=bind",
+	"workspaceFolder": "/root/src/TestSlide",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-azuretools.vscode-docker",
+				"ms-vscode.makefile-tools",
+				"ms-python.python"
+			]
+		}
+	}
+}


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM debian
+RUN apt update && apt -y install python3 python3-venv python3-dev make gcc
```
fornellas, line 2: I reckon the idea here is great, but it seems at this point, we'd have 3 paths for build:

- The "official" one at GH actions https://github.com/facebook/TestSlide/blob/main/.github/workflows/build.yml
- The already existing Dockerfile https://github.com/facebook/TestSlide/blob/main/Dockerfile
- The new Dockerfile added by this PR.

It looks like we'd benefit from having a single way of building the project.

The GH Action ("official") is based on Ubuntu-latest, and simply install the required python version and virtualenv. Perhaps, if the logic for these 2 steps is moved to the Makefile (eg: use [pyenv](https://github.com/pyenv/pyenv)), then we could have:

- GH action simply calls the Makefile.
- A single Dockerfile, which also calls the Makefile.
- Both based on the same ubuntu version (not "latest").

It seems this would enable us to have more reproducible builds.

shish: I forget why I created a second Dockerfile, I think it was a good idea at some point in the _middle_ of my experimentation, but it isn't needed - I've updated the devcontainer config to use the existing Dockerfile

(I agree with the other points, but think they belong in separate PRs, I might do those later if I still have time at the end of the hack week)

PR #360, opened 2024-03-25
[ci] also test in 3.11 / 3.12

shish:
[ci] also test in 3.11 / 3.12

3.12 is the current python stable version, we should probably make sure it works?


fornellas:
It looks like https://github.com/facebook/TestSlide/commit/2a18d5ec77e3e4898e44d0f0afbed6456ec4d58b broke the build a while back @macisamuele (not sure how this was allowed to be merged...).

